PR #107, opened 2024-10-15
refactor(web): Simplify <DataForm.Provider /> side effects

Cwejman:
By creating the withLobotomiser hoc we can let the provider start fresh when ever the mode changes. This heavyly simplifies the internal bussiness logic. We no longer need to try juggle the state and props according to mode changes. Just instanciate the compoent from scratch everytime.

Cool.

### DIFF
--- a/web/src/components/DataForm/DataFormDatePicker.tsx
+++ b/web/src/components/DataForm/DataFormDatePicker.tsx
@@ -21,7 +21,7 @@ export const DatePicker = ({
   fromDate,
 }: DataFormDateProps) => {
   const [isOpen, setIsOpen] = useState(false);
-  const { isSkeleton, isDisabled, body, setBodyProp } = useDataForm();
+  const { isLoading, isDisabled, body, setBodyProp } = useDataForm();
 
   return (
     <Field label={label} dataKey={dataKey} description={description}>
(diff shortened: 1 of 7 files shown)


Daviduber69:
noticed nothing out of the ordinary, everything seems to work fine. 
Did not see any Toast message when having 2 tabs open of the same page and editing a Ticket in 1 , and then checking the other one, but the branch isn't up to date with main so i dont know.
But the UI re-renders on the other tab without having to refreshing the site, so that's good.

I'm not sure if this is something that is or is not intended. But whenever a Vehicle doens't have an active Ticket, with a Status, you cannot 
edit that Vehicle ("Status is requried" tooltip on Save button when in Edit mode). Why would you not want the option to edit a Vehicle without an active status on a Ticket?
This is probably not relevant to this refactoring either, but i'm just writing it down here.

Something else that isn't relevant to this that i'm still writing down is this:
image

Should Hjulverkstan really be able to Rent out the same Vehicles, on the same date, X number of times?

Might be annoying to note this, but maybe Vehicles that are Rented out on a date interval should be disabled for renting out on those dates?

Again this branch isn't up to date and rebased with Main, so I might be revising this comment after that is done.

Cwejman:
U have tested locally and the "there has been changes to your at a do you want to refresh" toast does work, one just has to switch from chrome to another program and back for it to refresh.

Daviduber69:
I am approving this PR testing testing

Daviduber69 (approved):
I am once again approving this for real this time

PR #127, opened 2024-10-23
fix(web): Sorting on columns not working on admin pages

Daviduber69:
Sorting now works on the admin pages columns using the correct keys in useColumns.

### DIFF
--- a/web/src/data/location/queries.ts
+++ b/web/src/data/location/queries.ts
@@ -9,7 +9,14 @@ import * as api from './api';
 //
 
 export const useLocationsQ = () =>
-  useQuery<Location[], StandardError>(api.createGetLocations());
+  useQuery<Location[], StandardError>({
+    ...api.createGetLocations(),
+    select: (locations) =>
+      locations?.map((location) => ({
+        ...location,
+        vehicleCount: location.vehicleIds?.length || 0,
+      })),
+  });
 
 export interface UseLocationQProps {
   id: string;
--- a/web/src/root/Portal/PortalAdminEmployees/useColumns.tsx
+++ b/web/src/root/Portal/PortalAdminEmployees/useColumns.tsx
@@ -12,7 +12,7 @@ export default function useColumns() {
     () =>
       [
         {
-          key: 'name',
+          key: 'firstName',
           name: 'Name',
           renderFn: ({ firstName, lastName }) => (
             <IconLabel label={`${firstName} ${lastName}`} />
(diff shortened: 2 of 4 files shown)


web/src/data/location/api.ts
```diff
@@ -49,13 +49,15 @@ const transformBody = ({
   address,
   name,
   comment,
+  vehicleCount,
```
Cwejman, line 52: Why send vehicle count to the backend when editing / creating?

web/src/data/location/queries.ts
```diff
@@ -9,7 +9,17 @@ import * as api from './api';
...
 
 export const useLocationsQ = () =>
-  useQuery<Location[], StandardError>(api.createGetLocations());
+  useQuery<Location[], StandardError>({
+    ...api.createGetLocations(),
+    select: (locations) => {
+      return locations
+        .map((location) => ({
+          ...location,
+          vehicleCount: location.vehicleIds?.length || 0,
```
Cwejman, line 18: Vehicle count is not used anywhere?

Cwejman: And if a feature should be a separate commit from the fix

Daviduber69:
Hello boys i think i resolved the issues. 2 commits now , 1 feat 1 fix, no backend api nonsense . Lets go

Madasa-cell (approved):
This seems to be good and we can now sort on Admin pages by their columns. Everything that I tested worked and I found all the pages from the page I pulled down. And the code seems to be good.

web/src/data/location/queries.ts
```diff
@@ -9,7 +9,17 @@ import * as api from './api';
...
+        }))
+        .sort((a, b) => b.vehicleCount - a.vehicleCount);
```
Cwejman, line 20: Why sort here, when the user can sort by pressing each column?

Our of coherency I believe no sorting but the default order from the backend should be present unless manually sorted by the user.

And the backends sorting should be by created at I believe

Cwejman: Otherwise no comments

Cwejman: Well actually one comment, you've decided to split your work into two commits, but the feature commit is not really a feature for the user is it? It is rather a fix because without a vehicleCount field the sorting for that column did not work? Perhaps better to just do one fix commit

Daviduber69: It was much easier than I made it out to be and it should be sorted out now hehehe

PR #128, opened 2024-10-24
feat(web): Status changes on repair tickets

daniel-svendsen:
Adds new automatical status changes when changing ticket statuses depending on if customer owned or not.

issue#76

### DIFF
--- a/main/src/main/java/se/hjulverkstan/main/service/TicketServiceImpl.java
+++ b/main/src/main/java/se/hjulverkstan/main/service/TicketServiceImpl.java
@@ -170,6 +170,16 @@ public class TicketServiceImpl implements TicketService {
         customer.getTickets().add(ticket);
         ticket.setCustomer(customer);
 
+        // Update vehicle status to UNAVAILABLE if it's a repair ticket, not closed and not customer owned
+        if (ticket instanceof TicketRepair && ticket.getTicketStatus() != TicketStatus.CLOSED) {
+            for (Vehicle vehicle : vehicles) {
+                if (vehicle.getVehicleStatus() == VehicleStatus.AVAILABLE && !vehicle.isCustomerOwned()) {
+                    vehicle.setVehicleStatus(VehicleStatus.UNAVAILABLE);
+                    vehicleRepository.save(vehicle);
+                }
+            }
+        }
+
         ticketRepository.save(ticket);
         return convertToDto(ticket);
     }
@@ -179,7 +189,7 @@ public class TicketServiceImpl implements TicketService {
         Ticket ticket = ticketRepository.findById(id)
                 .orElseThrow(() -> new ElementNotFoundException("Ticket with id " + id + " not found"));
 
-        /*
+                /*
         TODO: Implement proper validation strategy
          Not sure what pattern we should be using here.
          Ways that validation could be done:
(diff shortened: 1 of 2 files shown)


Daviduber69 (approved):
Sweet

main/src/main/java/se/hjulverkstan/main/service/TicketServiceImpl.java
```diff
@@ -189,24 +199,52 @@ public TicketDto updateTicketStatus(Long id, TicketStatusDto ticketStatusDto) {
...
+                vehicleRepository.save(vehicle);
+            } else {
+                if (isRentTicket) {
+                    if (newStatus == TicketStatus.IN_PROGRESS) {
+                        if (vehicle.getVehicleStatus() == VehicleStatus.AVAILABLE) {
+                            vehicle.setVehicleStatus(VehicleStatus.UNAVAILABLE);
+                            vehicleRepository.save(vehicle);
+                        }
+                    } else if (newStatus == TicketStatus.CLOSED) {
+                        vehicle.setVehicleStatus(VehicleStatus.AVAILABLE);
```
Cwejman, line 224: I did not think that bikes become automatically available but that the user gets to set the new status in a front end dialog instead?

main/src/main/java/se/hjulverkstan/main/service/TicketServiceImpl.java
```diff
@@ -189,24 +199,52 @@ public TicketDto updateTicketStatus(Long id, TicketStatusDto ticketStatusDto) {
...
+                        vehicleRepository.save(vehicle);
+                    }
+                } else if (isRepairTicket) {
+                    if (newStatus == TicketStatus.CLOSED) {
+                        vehicle.setVehicleStatus(VehicleStatus.AVAILABLE);
```
Cwejman, line 229: Same here about automatically setting available

web/src/root/Portal/PortalShopTickets/ShopTicketsActions.tsx
```diff
@@ -93,9 +93,18 @@ export default function ShopTicketsActions({
               id: res.id,
             }),
           );
+          const notCustomerOwnedVehicles = vehicles.filter(
```
Cwejman, line 96: Better filter on the field isCustomerOwned?

Madasa-cell:
Currently if a Customer Owned vehicle and a Hjulverkstan Owned vehicle share the same Repair Ticket you are given the illusion of a choice for the Customer Owned vehicle that you can select anything when in fact, no matter what you pick the end result is always Archived. But maybe users of the portal shouldn't be able to place vehicles with different owners on the same ticket. I'm not to familiar with how they are supposed to work.

And then there is of course the problem with available which Cwejman has already mentioned.

Madasa-cell (commented):
Currently if a Customer Owned vehicle and a Hjulverkstan Owned vehicle share the same Repair Ticket you are given the illusion of a choice for the Customer Owned vehicle that you can select anything when in fact, no matter what you pick the end result is always Archived. But maybe users of the portal shouldn't be able to place vehicles with different owners on the same ticket. I'm not to familiar with how they are supposed to work.

And then there is of course the problem with available which Cwejman has already mentioned.

web/src/root/Portal/PortalShopTickets/ShopTicketsActions.tsx
daniel-svendsen: yes, pushed update

web/src/root/Portal/PortalShopTickets/ShopTicketsActions.tsx
```diff
@@ -93,9 +93,18 @@ export default function ShopTicketsActions({
...
+            (vehicle) =>
+              vehicle.vehicleStatus !== VehicleStatus.ARCHIVED &&
+              vehicle.vehicleStatus !== null &&
+              vehicle.vehicleStatus !== undefined,
+          );
+
           if (
             newStatus === TicketStatus.CLOSED &&
-            ticket.ticketType === TicketType.RENT
```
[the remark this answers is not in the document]
daniel-svendsen: believe i fixed. definetly pushed

web/src/root/Portal/PortalShopTickets/ShopTicketsActions.tsx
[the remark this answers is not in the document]
daniel-svendsen: i´ll leave it for now if not deemed prio for April event. can make an issue though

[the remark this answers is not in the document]
daniel-svendsen: done a seperate commit

web/src/root/Portal/PortalShopTickets/ShopTicketsActions.tsx
```diff
@@ -40,19 +39,15 @@ export default function ShopTicketsActions({
...
-  const vehicleQueries = useQueries({
-    queries: ticket.vehicleIds.map((id) => {
-      const getVehicleQuery = createGetVehicle({ id });
-      return {
-        queryKey: getVehicleQuery.queryKey,
-        queryFn: getVehicleQuery.queryFn,
-      };
-    }),
+  const getVehicles = createGetVehicles();
+  const { data: allVehicles } = useQuery({
```
Cwejman, line 43: Remember that we define our query hooks in the respective query.ts files in the data directory, and when we do we do so like

```typescript
export const useVehiclesQ = () => useQuery(api.createGetVehicles());
```
But this is of course already done and can be used instead like so:

```typescript
const vehiclesQ = useTicketsQ();
```

Cwejman:
> Currently if a Customer Owned vehicle and a Hjulverkstan Owned vehicle share the same Repair Ticket you are given the illusion of a choice for the Customer Owned vehicle that you can select anything when in fact, no matter what you pick the end result is always Archived. But maybe users of the portal shouldn't be able to place vehicles with different owners on the same ticket. I'm not to familiar with how they are supposed to work.
> 
> And then there is of course the problem with available which Cwejman has already mentioned.

Is this your conclusion from testing locally?

Looking at the code the dialog should only contain non customer owned vehicles. And the other, by the backends code, should be automatically archived.

web/src/root/Portal/PortalShopTickets/ShopTicketsActions.tsx
daniel-svendsen: done!

daniel-svendsen:
> > Currently if a Customer Owned vehicle and a Hjulverkstan Owned vehicle share the same Repair Ticket you are given the illusion of a choice for the Customer Owned vehicle that you can select anything when in fact, no matter what you pick the end result is always Archived. But maybe users of the portal shouldn't be able to place vehicles with different owners on the same ticket. I'm not to familiar with how they are supposed to work.
> > And then there is of course the problem with available which Cwejman has already mentioned.
> 
> Is this your conclusion from testing locally?
> 
> Looking at the code the dialog should only contain non customer owned vehicles. And the other, by the backends code, should be automatically archived.

it was indeed so. i´ve changed it so it only takes in notCustomedOwnedVehicles, and customerowned become automatically archived when closed

Cwejman:
Great, lets merge

Issue #123, opened 2024-10-23
Crash when creating- then deleting ticket and then navigating to inventory page


Daviduber69:

daniel-svendsen:
I could not reproduce this, closing.

daniel-svendsen:
![Image](https://github.com/user-attachments/assets/0ce7fc56-f56a-4ee4-b49c-f0eaded1daee)

Daviduber69:
### Potential causes (perhaps)
React Query cache not updating properly and might be referencing outdated or incorrect data after a ticket is deleted.

 ```
 const badges = ticketIds.map((ticketId) => {
    const { customerId, employeeId } = ticketsQ.data.find(
      (t) => t.id === ticketId,
    )!;

    const ticketEnum = ticketEnumsQ.data.find((e) => e.value === ticketId)!;
    const customerEnum = customerEnumsQ.data.find(
      (e) => e.value === customerId,
    )!;
    const employeeEnum = employeeEnumsQ.data.find(
      (e) => e.value === employeeId,
    )!;
```

From PortalShopTickets/useColumns.tsx (as the console error refers to), and look into the different queires in data/ticket/queries.ts
And the issue might be found there.

```
export const useTicketsAsEnumsQ = ({ dataKey = 'ticketId' } = {}) =>
  useQuery<Ticket[], StandardError, EnumAttributes[]>({
    ...api.createGetTickets(),
    select: (tickets): EnumAttributes[] =>
      tickets?.map((ticket) => ({
        dataKey,
        icon: enums.find(ticket.ticketType).icon,
        label: `#${ticket.id}`,
        value: ticket.id,
        ...(ticket.ticketStatus && {
          variant: {
            [TicketStatus.CLOSED]: 'outline',
            [TicketStatus.IN_PROGRESS]: 'success',
            [TicketStatus.READY]: 'warn',
            [TicketStatus.COMPLETE]: 'success',
          }[ticket.ticketStatus] as EnumAttributes['variant'],
        }),
      })) ?? [],
  });
```

e.g. maybe here.

Cwejman:
Analysis seems correct, the code was written, expecting a ticket to always be there. I think I could not think of a reason for it not to be at the time of writing and left it like so so that if I was wrong it would blow up in our faces.

One way is to solve it by not showing badge if there is not a ticket for it.

[...]

PR #153, opened 2025-02-14
fix(web): Crash on badges when ticketId is not found

Daviduber69:
#123 Now filtering to find all valid ticketIds before mapping


daniel-svendsen (commented):
seems to be working now!

Cwejman (approved):
Looks good, simple fix and good enough commit name

Cwejman:
> seems to be working now!

Remember to also approve in the UI so that the PR is unblocked, overriding with admin privileges for now

Issue #48, opened 2024-09-04
Cant see the users on admin page


Christopher-Java:
Its possible to create users and log in as that user, however for some reason its not possible to see the users on the page.

-------------------

## Frontend

Seems like this is a Backend Issue.

## Backend

The problem here seems to be that Users does not have the property "createdAt" in its endpoint which is necessary for it to be rendered in the columns. To fix this it needs to implement Auditable that handles when somethings is created (createdAt) or edited (editedAt) and who it was that did it (createdBy / editedBy).

## Console Error

```typescript
2025-02-05T10:03:57.236+01:00  WARN 17456 --- [nio-8080-exec-9] .m.m.a.ExceptionHandlerExceptionResolver : Resolved [org.springframework.data.mapping.PropertyReferenceException: No property 'createdAt' found for type 'User']
2025-02-05T10:03:57.262+01:00 DEBUG 17456 --- [nio-8080-exec-1] org.hibernate.SQL                        : select u1_0.id,u1_0.email,u1_0.password,u1_0.username from users u1_0 where u1_0.username=?
2025-02-05T10:03:57.264+01:00 DEBUG 17456 --- [nio-8080-exec-1] org.hibernate.SQL                        : select r1_0.user_id,r1_1.id,r1_1.name from user_roles r1_0 join roles r1_1 on r1_1.id=r1_0.role_id where r1_0.user_id=?
2025-02-05T10:03:57.270+01:00  WARN 17456 --- [nio-8080-exec-1] .m.m.a.ExceptionHandlerExceptionResolver : Resolved [org.springframework.data.mapping.PropertyReferenceException: No property 'createdAt' found for type 'User']

Internal Error : No property 'createdAt' found for type 'User'
```

daniel-svendsen:
![Image](https://github.com/user-attachments/assets/39b8790a-0b1a-48de-bb72-e3ceae1deac7)

this is what i get when adding:
  console.log('User API Endpoint:', endpoints.user);
  console.log('Axios headers:', instance.defaults.headers);
  console.log(localStorage.getItem('token'));

in PortalAdminUsers/index.tsx.

It seems to be something with endpoint + authentication

Cwejman:
Looks good and ready 👍

PR #156, opened 2025-02-17
fix(api): Enables rendering on Admin/Users

Madasa-cell:
Added logic for when a User is created or edited.

It is critical that users can see this in our design and for the backend to handle the data correctly.

Added Auditable to Users and other minor changes in other files to accomodate createdAt and editedAt.

-----------

Solves issue https://github.com/Hjulverkstan/hjulverkstan/issues/48.

### DIFF
--- a/main/src/main/java/se/hjulverkstan/main/dto/user/SignupRequest.java
+++ b/main/src/main/java/se/hjulverkstan/main/dto/user/SignupRequest.java
@@ -7,6 +7,7 @@ import lombok.NoArgsConstructor;
 import lombok.ToString;
 import se.hjulverkstan.main.model.ERole;
 
+import java.time.LocalDateTime;
 import java.util.Set;
 
 @Data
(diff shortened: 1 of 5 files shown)


daniel-svendsen (commented):
users now rendering!

main/src/main/java/se/hjulverkstan/main/dto/user/SignupRequest.java
```diff
@@ -32,4 +33,10 @@ public class SignupRequest {
     private String password;
 
     private Set<ERole> roles;
+
+    // Meta data
+    private LocalDateTime createdAt;
+    private LocalDateTime updatedAt;
```
Cwejman, line 39: Does SignupRequest really need the meta data?

Cwejman (commented):
One comment only about implementation. Bonus for writing a commit body message – though please rewrite the commit message to explain the issue and not solution. As previously discussed and mentioned in our git best practices

main/src/main/java/se/hjulverkstan/main/dto/user/SignupRequest.java
Madasa-cell: No it did not the meta data.

Cwejman:
Commit message still needs to be rewritten, here is a link to our git strategy regarding how git messages should be written: [git strategy](https://github.com/Hjulverkstan/hjulverkstan/blob/main/GUIDELINES.md#git-strategy-)

Cwejman (approved):

PR #157, opened 2025-02-17
fix(web): filtering tickets also includes timestamps

daniel-svendsen:
adds a util now used in fuzzyMatchfn that removes the timestamp from start_date and end_date since that was included when searching in the tickets page.

bug#70

### DIFF
--- a/web/src/root/Portal/PortalShopTickets/ShopTicketsFilters.tsx
+++ b/web/src/root/Portal/PortalShopTickets/ShopTicketsFilters.tsx
@@ -18,6 +18,7 @@ import {
   VehicleShortcutAction,
   VehicleShortcutLocationState,
 } from '../PortalShopInventory/ShopInventoryActions';
+import { matchDateWithoutTimestamp } from '@utils';
 
 export default function ShopTicketFilters() {
   const ticketsQ = useTicketsAggregatedQ();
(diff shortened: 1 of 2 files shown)


web/src/components/DataTable/DataTableFilterSearch.tsx
```diff
@@ -14,13 +14,24 @@ import {
...
 
 export const fuzzyMatchFn = (keys: string[], word: string, row: Row) =>
-  keys.some((key) => row[key]?.toLowerCase().includes(word));
+  keys.some((key) => {
+    let field = row[key];
+    // For start and enddate fields, remove the timestamp before matching since it filters for that aswell
+    if (
+      typeof field === 'string' &&
+      (key === 'startDate' || key === 'endDate')
+    ) {
```
Cwejman, line 30: I think it is better to write a util for matching a date without timestamp and using it in each page's matchFn. Thus keeping the fuzzy match fn general and focusing on composition of the matchFn for each page being modular rather then restricted.

We already compose various utilities in each page's matchFn for the search filter using pipe operators. Using a "match date only" function in this composition makes the most sense.

daniel-svendsen: updated and pushed

Cwejman (approved):

PR #163, opened 2025-02-19
Feature/customer owned vehicles unselectable on rentaltickets

Daviduber69:


web/src/root/Portal/PortalShopTickets/ShopTicketsFields.tsx
```diff
@@ -13,6 +13,13 @@ export default function ShopTicketFields() {
   const employeeEnumsQ = useEmployeesAsEnumsQ();
   const customerEnumsQ = useCustomersAsEnumsQ();
 
+  const filteredVehiclesForRentalTickets =
+    body.ticketType === TicketType.RENT
+      ? (vehicleEnumsQ.data ?? []).filter(
+          (vehicle) => vehicle.label.charAt(0) !== '#',
```
daniel-svendsen, line 19: im thinking its possible to get if its customerowned?

daniel-svendsen (approved):

web/src/root/Portal/PortalShopTickets/ShopTicketsFields.tsx
Cwejman: Agree filtering on if starts with '#' is a bit cheap and could lead to issues. There is a "truer" source and that is the `isCustomerOwned` field. That would mean a bit more complexity as that field is not available on enums. But I think a better solution none the less

Cwejman (changes requested):
Two comments regarding the backend work:

1. It seems that some changes are just about indentation. It is best to exclude any such work from the commit as it makes the commit larger that it in fact is. A note for the future – not the end of the world
2. Filtering on isCustomerOwned solves the issue at hand but not quite right. Could be connected to our GUIDELINES.md file where we talk about explocioty and that if something fails it should not fail silently. If a client send a list of vehicles and the response is a success then the client, if it did not thoroughly compare the response expect that the ordered change has been executed. It is better to validate and just throw an error that this action is not possible.

web/src/root/Portal/PortalShopTickets/ShopTicketsFields.tsx
Daviduber69: I removed the previously reformated code in TicketServiceImpl, which made it appear again as changed to its previous state but its still in this latest commit T_T but yeah, its back to the way it was, which was dumb of me

PR #164, opened 2025-02-19
feat(web): prevent status changes on open tickets

daniel-svendsen:
you now cannot change status on vehicles if that vehicle has an open ticket

issue#77

### DIFF
--- a/main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
+++ b/main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
@@ -128,7 +128,14 @@ public class VehicleServiceImpl implements VehicleService {
 
     @Override
     public VehicleDto editVehicleStatus(Long id, EditVehicleStatusDto newStatus) {
-        Vehicle selectedVehicle = vehicleRepository.findById(id).orElseThrow(() -> new ElementNotFoundException(ELEMENT_VEHICLE));
+        Vehicle selectedVehicle = vehicleRepository.findById(id)
+                .orElseThrow(() -> new ElementNotFoundException(ELEMENT_VEHICLE));
+
+        for (Ticket ticket : selectedVehicle.getTickets()) {
+            if (ticket.isOpen()) {
+                throw new UnsupportedVehicleStatusException("Vehicle status cannot be changed when it has an active ticket.");
+            }
+        }
 
         selectedVehicle.setVehicleStatus(newStatus.getNewStatus());
 
--- a/web/src/root/Portal/PortalShopInventory/ShopInventoryActions.tsx
+++ b/web/src/root/Portal/PortalShopInventory/ShopInventoryActions.tsx
@@ -18,6 +18,9 @@ import { PortalTableActionsProps } from '../PortalTable';
 import * as Tooltip from '@radix-ui/react-tooltip';
 import * as enums from '@data/vehicle/enums';
 
+import { useTicketsQ } from '@data/ticket/queries';
+import { TicketStatus } from '@data/ticket/types';
+
 export enum VehicleShortcutAction {
   CREATE_TICKET = 'CREATE_TICKET',
   FILTER_BY_VEHICLE = 'FILTER_BY_VEHICLE',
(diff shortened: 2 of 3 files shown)


main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
```diff
@@ -128,7 +128,20 @@ public EditVehicleDto editVehicle(Long id, EditVehicleDto editVehicle) {
...
+        Vehicle selectedVehicle = vehicleRepository.findById(id)
+                .orElseThrow(() -> new ElementNotFoundException(ELEMENT_VEHICLE));
+
+        for (Ticket ticket : selectedVehicle.getTickets()) {
+            if ((ticket.getTicketType() == TicketType.REPAIR &&
+                    (ticket.getTicketStatus() == TicketStatus.READY ||
+                            ticket.getTicketStatus() == TicketStatus.IN_PROGRESS ||
+                            ticket.getTicketStatus() == TicketStatus.COMPLETE)) ||
+                    (ticket.getTicketType() == TicketType.RENT &&
+                            (ticket.getTicketStatus() == TicketStatus.READY ||
```
Cwejman, line 140: Perhaps just if not null or not closed?

Also I think nicer if ticket just has a method on it .isOpen() that does this check inside it. Keep the service cleaner and more decoupled.

web/src/root/Portal/PortalShopInventory/ShopInventoryActions.tsx
```diff
@@ -109,25 +143,30 @@ export default function ShopInventoryActions({
...
-                disabled={hasTickets}
+        <DropdownMenu.Item
+          onClick={(e) => e.stopPropagation()}
+          onSelect={handleDeleteClick}
+          disabled={hasTickets}
+        >
+          Delete
+          <DropdownMenu.Shortcut>⌘⌫</DropdownMenu.Shortcut>
+        </DropdownMenu.Item>
+        {hasTickets && (
```
Cwejman, line 154: Has this been tested locally? Looks like the delete button no longer is connected to the tooltip and that the content has now become the trigger...

web/src/root/Portal/PortalShopInventory/ShopInventoryActions.tsx
```diff
@@ -99,6 +89,50 @@ export default function ShopInventoryActions({
...
+      ticket.ticketStatus !== undefined &&
+      [TicketStatus.READY, TicketStatus.IN_PROGRESS].includes(
+        ticket.ticketStatus,
+      ),
+  );
+
+  const statusChangeDisabled = hasActiveRepairTicket || hasActiveRentalTicket;
+  const tooltipMessage = hasActiveRepairTicket
+    ? 'Vehicle status cannot be changed when it has an active repair ticket.'
+    : hasActiveRentalTicket
```
Cwejman, line 132: Not sure there is a value in different messages for repair or rental. Complexity could therefore be reduced.

Also I'm not sure we need to differentiate between rental and repair to compare with different statuses. perhaps some(ticket => [READY, IN_PROGRESS; COMPLETE].includes(ticket.ticketStatus)) is enough?

Cwejman: And that is the same for the backend then. 

I proposed to ask not null and not closed instead there though. Not sure which is the best approach if there is a better one but both these options are simpler.

Or perhaps I'm missing something

Cwejman (changes requested):

web/src/root/Portal/PortalShopInventory/ShopInventoryActions.tsx
```diff
@@ -44,6 +47,17 @@ export default function ShopInventoryActions({
...
+  const ticketsQ = useTicketsQ();
+  const vehicleTickets =
+    ticketsQ.data?.filter((ticket) => vehicle.ticketIds.includes(ticket.id)) ??
+    [];
+  const hasActiveTicket = vehicleTickets.some(
+    (ticket) =>
+      ticket.ticketStatus !== undefined &&
+      ticket.ticketStatus !== TicketStatus.CLOSED,
+  );
+  const statusChangeDisabled = hasActiveTicket;
```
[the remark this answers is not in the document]
daniel-svendsen: fixed!

Cwejman (approved):

daniel-svendsen:
> The tooltip is showing when the tickets are ready, in progress or completed!
> 
> But i dont know if this is a big deal but the issue says that the message should be specific to the type of ticket image
> 
> But the message shown is the same for both types:
> 
> image
> 
> I don't personally mind that it is a general message for both types but i thought i might comment on it.

@Cwejman thought it would be ok with just one general, i did for both at first

Daviduber69 (approved):

Issue #152, opened 2025-02-13
Uploading images from computer in WebEdit


Daviduber69:
[#28](https://github.com/Hjulverkstan/hjulverkstan/issues/28) Edit this component and add props to that component so that in WebEdit you will be able to upload images from a computer and not only mobile.

The component right now only has a text saying "Use a mobile phone to upload images" while in Inventory, but in WebEdit it will have the option to upload images from the computer.

Implement a feature to attach the image when clicking on the Edit-button in WebEdit. The image will be sent to the backend, where it will go to the S3 cloud, and then backend will return an URL which will be stored in DataForm.

### Validation
Validate that the chosen file is the correct format (e.g. jpg, png, etc). And have the OS automatically filter on only image files when opening the upload window. And maybe have a set maximum file size added.

You should only be able to upload images on Hjulverkstan Owned Vehicles.

And the image dimensions maybe should be automatically be set to fit the page's set standard, or maybe have a maximum ceiling for pixels.

### We need to send the data to the backend 
that #28 will create an endpoint for. That endpoints will be `/v1/image/upload` for uploading image.
Look into **web/src/data/vehicles/**, and all the relevant files.

Needs a different design from the Inventory page and shoud look like this:

![Image](https://github.com/user-attachments/assets/644f85b3-3efe-4ec3-898c-823a2e7994d9)

This is sort of a universal reusable component so it should be usable within General/event/blogs.

Look into how uploading files from a PC works and implement that. (npm install react-uploader, perhaps?)

### Disable Save/edit button during image upload

```
// Adding something like this to the DataFormProvider proving a function to be passed thourgh context, to be consumed by the image component
const DataFormProvider = () => {
  const [manualIssuesMap, setManualIssueMap] = useState({});
  
  const registerManualIssue = (key: string, message: string | null) =>
    setManualIssueMap(map => {
      if (message) return ({ [key]: message });
      else {
        delete map[key];
        return map;
      }
    });
 
  // To me concatenated with bodyIssues from zod
  const manualIssues = Object.values(manualIssuesMap);
```
}

Issue #29, opened 2024-06-07
Image component for DataForm with delete functionality


Cwejman:
Perhaps the image upload and the DataForm component implementation is to be decoupled. Regardless a feature set along the lines of:

### Create a new component in component/DataForm  
and use it with **DataForm Provider** . (this is assuming we're not decoupling it)

Use this component in **ShopInventoryFields**. 
In Inventory on computer mode there will be a text saying "Use a mobile phone to upload images". See Jona's comment below for more info.

In Edit you should also be able to delete the image.
The Delete Image button should remove the url from the component, and maybe have a boolean switch to indicate that the Vehicle doesn't have an image.
Look into web/src/data/vehicles/, and all the relevant files.
And here when deleting the image in the Portal, we will send data to the Delete endpoint in backend.

**_mock code example:_**
```
<DataForm.Image
  dataKey=""
 label=""
/>
```

Attaching in image while creating a new vehicle might be hard to implement seeing as the image needs to have a vehicle ID which is only created when the vehicle already has been created.

---------------------------
### Visual representation of what it could and should look like below
Except that there should be no Upload Image ("Browse") thingy mcthing, just "Use mobile to upload images"
And you should see the uploaded image, and be able to delete the image.
<details><summary>El prototypo, señorita</summary>
<p>

![Image](https://github.com/user-attachments/assets/9e774b93-9c09-4413-9155-442f2281b06d)

</p>
</details>

Cwejman:
I dont see the need for most of the props in the pseudo code example

```typescript
<DataForm.Image
  dataKey=""
 label=""
/>
```

Should only be needed I think.

Cwejman:
Should be clear that this task includes adding this component to the inventory form so that it is actually used.

Daviduber69:
> When an image is being uploaded we need to prevent the save/create button. How do we do that? We dont want to add image specific logic to the dataform provider unless absolutely needed. Do we need to provide some kind of api through context, i.e. useDataForm() for the image component to use to prevent save/edit?

Can it be done by populating the body url with a placeholder value like “UPLOADING” and the form validation shows an error if UPLOADING? We dont want the error to be shown next to the input though, only on the button, so in the case we need to build on error messages from zod so that they can be save button only / button and error banner next to the input field.

jimmsi:
When the UI handles image loading failures, it’s crucial to differentiate between two scenarios:

Failed Image Loading (Temporary Error):
[...]

PR #166, opened 2025-02-24
feat(web): Implementing image component with delete and upload functionality

Daviduber69:
#29 and #152  

Upload and delete functionality for images in a brand new ImageComponent

(1 commit instead of 2 now becaues this is the way)

### DIFF
--- a/web/src/components/DataForm/DataFormImage.tsx
+++ b/web/src/components/DataForm/DataFormImage.tsx
@@ -1,0 +1,182 @@
+import { Field, FieldProps, Mode, useDataForm } from './';
+import { useUploadImageM } from '@data/image/mutations';
+import React, { useEffect, useRef, useState } from 'react';
+import Error from '@components/Error';
+import Spinner from '@components/Spinner';
+import { ImageUpIcon, Trash2Icon } from 'lucide-react';
+import { endpoints } from '@data/api';
+import { IconButton } from '@components/shadcn/Button';
+
+export interface ImageProps extends Omit<FieldProps, 'children'> {
+  disableImageUpload?: string;
+}
+
+export const Image = ({ dataKey, label, disableImageUpload }: ImageProps) => {
+  const { body, setBodyProp, mode, registerManualIssue } = useDataForm();
+  const uploadImageM = useUploadImageM();
+
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const [imageError, setImageError] = useState(false);
+
+  const handleDelete = () => {
+    setBodyProp('tempDeleteImageURL', body[dataKey]);
+    setBodyProp(dataKey, null);
+  };
+
+  const handleImageUpload = async (
+    event:
+      | React.ChangeEvent<HTMLInputElement>
+      | React.DragEvent<HTMLDivElement>,
+  ) => {
+    const file =
+      'files' in event.target
+        ? event.target.files?.[0]
+        : (event as React.DragEvent<HTMLDivElement>).dataTransfer.files[0];
+
+    if (!file) return;
+
+    registerManualIssue('image', 'Cannot save while image is uploading.');
+
+    uploadImageM
+      .mutateAsync({ file })
+      .then((imageURL) => {
+        setBodyProp(dataKey, imageURL);
+      })
+      .finally(() => {
+        registerManualIssue('image', null);
+      });
+  };
+
+  useEffect(() => {
+    setImageError(false);
+  }, [body[dataKey]]);
+
+  const handleError = () => {
+    setImageError(true);
+  };
+
+  const handleUploadClick = () => {
+    fileInputRef.current?.click();
+  };
+
+  const handleDragEvent = (event: React.DragEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    event.stopPropagation();
+  };
+
+  const handleDropEvent = async (event: React.DragEvent<HTMLDivElement>) => {
+    handleDragEvent(event);
+    await handleImageUpload(event);
+  };
+
+  return (
+    <Field label={label} dataKey={dataKey}>
+      <div
+        className={`relative flex h-[200px] max-w-[500px] flex-col items-center
+        justify-center gap-2 overflow-hidden rounded-lg border ${
+          !body[dataKey] ? 'border-dashed bg-gray-100' : ''
+        }`}
+      >
+        {mode === Mode.EDIT && body[dataKey] && (
+          <IconButton
+            variant="destructive"
+            className="absolute right-2 top-2 z-10"
+            icon={Trash2Icon}
+            onClick={handleDelete}
+          />
+        )}
+        {uploadImageM.isPending ? (
+          <div
+            className="flex h-8 w-8 items-center justify-center rounded-full
+              bg-gray-200 p-2"
+          >
+            <Spinner className="h-5 w-5 flex-shrink-0 " visible />
+          </div>
+        ) : (
+          <>
+            {!disableImageUpload && mode === Mode.EDIT && body[dataKey] && (
+              <>
+                <IconButton
+                  variant="outline"
+                  className="absolute right-12 top-2 z-10"
+                  icon={ImageUpIcon}
+                  onClick={handleUploadClick}
+                />
+                <input
+                  ref={fileInputRef}
+                  type="file"
+                  id={`file-upload-${dataKey}`}
+                  accept="image/*"
+                  onChange={handleImageUpload}
+                  className="hidden"
+                />
+              </>
+            )}
+            {body[dataKey] ? (
+              <div
+                className="relative flex h-full w-full items-center
+                  justify-center bg-gray-100"
+              >
+                {!imageError ? (
+                  <img
+                    src={body[dataKey]}
+                    alt="Uploaded"
+                    className="h-full w-full object-cover"
+                    onError={handleError}
+                  />
+                ) : (
+                  <Error
+                    error={{ error: 'NOT_FOUND', endpoint: endpoints.image }}
+                  />
+                )}
+              </div>
+            ) : disableImageUpload ? (
+              <p
+                className="absolute flex h-full w-full items-center
+                  justify-center text-center text-sm text-gray-500"
+              >
+                {disableImageUpload}
+              </p>
+            ) : (
+              !disableImageUpload &&
+              mode === Mode.EDIT && (
+                <div
+                  className="relative flex h-full w-full flex-col items-center
+                    justify-center text-center"
+                  onDragOver={handleDragEvent}
+                  onDragEnter={handleDragEvent}
+                  onDragLeave={handleDragEvent}
+                  onDrop={handleDropEvent}
+                >
+                  <input
+                    type="file"
+                    id={`file-upload-${dataKey}`}
+                    accept="image/*"
+                    onChange={handleImageUpload}
+                    className="hidden"
+                  />
+                  <div
+                    className="flex h-8 w-8 items-center justify-center
+                      rounded-full bg-gray-200 p-2"
+                  >
+                    <ImageUpIcon className="h-5 w-5 text-gray-600" />
+                  </div>
+                  <p className="text-sm text-gray-500">Drag images here, or</p>
+                  <label
+                    htmlFor={`file-upload-${dataKey}`}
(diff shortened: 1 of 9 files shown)


Daviduber69:
I didnt use shadcn/ <Button /> for this since it was not possible to adjust the <TrashIcon> inside that <Button>, for the trashicon to be bigger I had to make the button very large, and when the button was a reasonable size, the TrashIcon was reaaally small, so I used a div instead.

web/src/root/Portal/PortalShopInventory/ShopInventoryFields.tsx
```diff
@@ -11,6 +11,14 @@ export default function ShopInventoryFields() {
 
   return (
     <>
+      {(mode === Mode.EDIT || mode === Mode.READ) && (
+        <DataForm.Image
+          label="Image"
+          dataKey="vehicleImage"
+          disableImageUpload="Open portal on a mobile device to upload images."
+        />
+      )}
```
Daviduber69, line 20: Uploading images shouldnt be possible in Read, I will sort this out later

jimmsi:
I think we should use the existing property imageURL on vehicle. Try to use it instead and we'll see how it turns out and how to address potential issues.

jimmsi (commented):
I think we should use the existing property imageURL on vehicle. Try to use it instead and we'll see how it turns out and how to address potential issues.

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export const Image = ({ dataKey, label, disableImageUpload }: ImageProps) => {
+  const { body, setBodyProp, mode, registerManualIssue } = useDataForm();
+  const [isUploading, setIsUploading] = useState(false);
+  const deleteImageM = useDeleteVehicleImageM();
+  const uploadImageM = useCreateImageM();
+
+  const handleDelete = () => {
+    setBodyProp(dataKey, '');
```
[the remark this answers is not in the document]
Cwejman: After discussions over video-meet we've concluded that the delete should be scheduled on the save action:

like so (in index.tsx PortalForm props)

```typescript
 saveMutation={({ tempDeleteImgUrl, ...body }) => tempDeleteImgUrl
                ? deleteImageM(tempDeleteImgURL).then(() => editVehicleM.mutateAsync(body))
                : editVehicleM.mutateAsync(body)}
```

And the delete handler can then just
```typescript
  const handleDelete = () => {
    setBodyProp('tempDeleteImageUrl', body[dataKey])
    setBodyProp(dataKey, null);
  };
```

web/src/components/DataForm/DataFormProvider.tsx
```diff
@@ -94,15 +95,32 @@ export const Provider = U.withLobotomizer(
...
+
+    // To me concatenated with bodyIssues from zod
+    const manualIssues = Object.values(manualIssuesMap);
+
     const bodyIssues = useMemo(() => {
       const zodRet = zodSchema.safeParse(body);
       return zodRet.success ? [] : zodRet.error.issues;
     }, [body]);
 
+    const allIssues = [...bodyIssues, ...manualIssues];
```
Cwejman, line 122: I guess you need to convert the array if just message strings to the same format as body issues, I think they are objects no?

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,122 @@
...
+      onError: (error) => {
+        console.log(error);
+        setIsUploading(false);
+      },
+    });
+  };
+
+  const handleSaveButtonWhileUploadingImage = () => {
+    if (isUploading) {
+      registerManualIssue('image', 'Cannot save while image is uploading.');
```
Cwejman, line 49: Unnecessary to decouple into a function just include the if statement where it is invoked

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const file = event.target.files?.[0];
+    if (!file) return;
+
+    setIsUploading(true);
+    handleSaveButtonWhileUploadingImage();
+
+    await uploadImageM.mutateAsync(file, {
+      onSuccess: (imageURL) => {
+        setBodyProp(dataKey, imageURL);
+        setIsUploading(false);
```
Cwejman, line 38: Unsuccessful remove the manual issue by registering it with a false value like `null`

Cwejman (changes requested):

Cwejman:
Converted to draft until Jimmiy's branch is closed

Cwejman:
Undrafting as jimmy's branch is closed – I'll take a look at the design when at the office tomorrow – are you planning to come in or do you need to work from home?

web/src/components/DataForm/DataFormImage.tsx
Cwejman, line 1: There is no loading indicator when uploading

Cwejman, line 1: Screenshot 2025-03-04 at 14 02 30

Trash button also does not look like mockup

Cwejman (changes requested):

Daviduber69:
(latest force push is a WIP)

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -31,59 +32,79 @@ export const Image = ({ dataKey, label, disableImageUpload }: ImageProps) => {
...
         registerManualIssue('image', null);
       });
   };
 
-  const handleImageError = () => {
-    setIsImageLoaded(false);
-  };
   const isValidUrl = (url: string) => {
-    return /^https:\/\//.test(url);
+    return /^https:\/\/.*hjulverkstan.*\./.test(url);
```
Daviduber69, line 42: the reason for this is that with faulty urls in img src, like there is now in the seed file, its still an URL, so I added validation for our URL seeing as this is the url we're gonna be using for uploading images

So i couldnt simply check if there was an URL present or not

Madasa-cell (commented):
Here are some basic critiques
1. If a file that is being uploaded is larger than 1MB (such as 1.1MB file) then you'll need to use these for that. Might be a problem.
   1a. spring.servlet.multipart.max-file-size=10MB
   1b. spring.servlet.multipart.max-request-size=10MB
2. When you delete an image you can't access the buttons either add or delete images.
3. You can delete the placeholder image, maybe disable the delete button if it is the placeholder image? Yeah, according to the example image this and the image swap button shouldn't be here yet.
4. You cannot drag and drop images. But that isn't something the task says you had to do, its just a design thing you had to do.
   4a. Well actually, maybe by implication he wants that functionality? Due to the example images.
5. You can upload images on any type of vehicle ownership, not just those owned by Hjulverkstan.

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -31,59 +33,77 @@ export const Image = ({ dataKey, label, disableImageUpload }: ImageProps) => {
...
+  };
+
+  const handleError = () => {
+    if (uploadImageM.error || !isValidUrl(body[dataKey])) {
+      return (
+        <Error error={{ error: 'NOT_FOUND', endpoint: endpoints.image }} />
+      );
+    }
+  };
+
```
jimmsi, line 53: We have two distinct scenarios regarding image loading errors:

Valid URL, but image missing or fails to load:
The URL passes the validation (the URL matches our bucket name, e.g. includes hjulverkstan), but the actual image resource does not exist, cannot load, or is otherwise corrupted.

Invalid URL:
The URL fails the validation check (isValidUrl), indicating it's incorrectly formatted or pointing to the wrong bucket.

For scenario 2 (Invalid URL), the handleError function is called explicitly, which successfully triggers rendering of the <Error /> component. For scenario 1 (Valid URL, but missing image), React's built-in onError event triggers handleError. However, currently, this does not properly display an error message due to React ignoring JSX returned from event handlers. This results in no visible feedback to the user.

Of course, we can adjust scenario 2 so that the same error message is displayed. However, it's worth considering if we actually want to handle these as two distinct errors, each communicated to the user with a separate error message.

Possible message 

1. Image could not be loaded. (and maybe also: Please try uploading the image again.)
2. The provided image URL is invalid. (and maybe also: Please try uploading a new image.)

daniel-svendsen:
tried with rebasing onto my work and got it working with delete. 
not much to say about the implementation or design besides from what i can see is good.

web/src/components/DataForm/DataFormImage.tsx
jimmsi: Since we found another solution, we can ignore the above comment for now...

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,178 @@
...
+      | React.ChangeEvent<HTMLInputElement>
+      | React.DragEvent<HTMLDivElement>,
+  ) => {
+    const file =
+      'files' in event.target
+        ? event.target.files?.[0]
+        : (event as React.DragEvent<HTMLDivElement>).dataTransfer.files[0];
+
+    if (!file) return;
+    if (uploadImageM.isPending) {
```
Cwejman, line 35: Should one be able to upload an image while it is loading?


Why?

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,178 @@
...
+    <Field label={label} dataKey={dataKey}>
+      <div
+        className={`relative flex h-[200px] max-w-[500px] flex-col items-center
+        justify-center gap-2 overflow-hidden rounded-lg border ${
+          !body[dataKey] ? 'border-dashed bg-gray-100' : ''
+        }`}
+      >
+        {mode === Mode.EDIT && body[dataKey] && (
+          <IconButton
+            className="absolute right-2 top-2 z-10 bg-red-600"
```
Cwejman, line 81: Create a variant instead, or perhaps there is actually already one to use. The current styling is not sufficient, on hover it becomes pink.

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,178 @@
...
+            icon={Trash2Icon}
+            onClick={handleDelete}
+          />
+        )}
+        {!disableImageUpload && mode === Mode.EDIT && body[dataKey] && (
+          <>
+            <IconButton
+              className="absolute right-12 top-2 z-10 bg-white text-gray-600"
```
Cwejman, line 89: Same with this style

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,178 @@
...
+            ) : (
+              <Error
+                error={{ error: 'NOT_FOUND', endpoint: endpoints.image }}
+              />
+            )}
+          </div>
+        ) : disableImageUpload ? (
+          <p
+            className="absolute flex h-full w-full items-center justify-center
+              text-center text-gray-500"
```
Cwejman, line 124: I think text should be the same size as other all other components for dataform, error.tsx included. They use text-sm

web/src/components/DataForm/DataFormImage.tsx
```diff
@@ -0,0 +1,178 @@
...
+                      rounded-full bg-gray-100 p-2"
+                  >
+                    <ImageUpIcon className="h-5 w-5 text-gray-600" />
+                  </div>
+                  <p className="text-gray-500">Drag images here, or</p>
+                  <label
+                    htmlFor={`file-upload-${dataKey}`}
+                    className="cursor-pointer font-bold text-pink-600"
+                  >
+                    browse
```
Cwejman, line 166: This should have some kind of alternate style on hover since it is a clickable. For instance opacity.

Cwejman (changes requested):
Here are a few comments.

Also the spacing is a bit irregular, that is code-wise. Some parts have empty lines between them and others are pushed together when perhaps the would be better treated as a new section. But I won't specify where, I dont really feel like I can be that nitty-gritty so rather just a hint, does not need solving.

web/src/components/DataForm/DataFormImage.tsx
Cwejman, line 1: Oh, also, the dataform still flashes a bit when loading. It shows browse for a brief moment when it should not.

Do check if this is from you, if handle the loading states right or if we have a bug somewhere else

Cwejman (approved):

Issue #143, opened 2025-02-03
Unable to create new employees in Admin portal


Daviduber69:
Unable to create new employees via the form on the Admin portal.

```
endpoint: '/employee', status: 400, error: 'UNKNOWN_ERROR', message: '[field: lastName, errors: Last name is required, f…field: firstName, errors: First name is required]', refreshSuccess: undefined}
endpoint
: 
"/employee"
error
: 
"UNKNOWN_ERROR"
message
: 
"[field: lastName, errors: Last name is required, field: email, errors: Email is required, field: employeeNumber, errors: Employee number is required, field: personalIdentityNumber, errors: Personal Identity Number is required, field: firstName, errors: First name is required]"
refreshSuccess
: 
undefined
status
: 
400
[[Prototype]]
: 
Object
```

Daviduber69:
Just wanted to add that you can as of now delete an employee and if that employee is bound to a ticket/vehicle the portal will crash.

Internal Error : Cannot invoke "se.hjulverkstan.main.model.Employee.getId()" because the return value of "se.hjulverkstan.main.model.Ticket.getEmployee()" is null

Which should be disabled

PR #174, opened 2025-03-03
fix: Unable to create new employees in Admin portal

jimmsi:
Fixes the issue preventing new employees from being created in the admin portal, removes the `employeeNumber` property from both the backend and frontend to align with the current data structure. Employee properties in the frontend now correctly match the backend. It makes the personal number and comment fields optional for both customers and employees. Backend validation for personal numbers now matches the frontend implementation. It also fixes a bug where fields that are edited and then cleared are sent as empty strings instead of `undefined`. Additionally, it allows personal number /age field to be missing for both customer and employee.

Solves #143


web/src/data/customer/form.ts
```diff
@@ -23,6 +23,8 @@ export const customerZ = z.discriminatedUnion('customerType', [
   }),
   customerBaseZ.extend({
     customerType: z.literal(CustomerType.PERSON),
-    personalIdentityNumber: swedishPIN,
+    personalIdentityNumber: swedishPIN.transform((val) =>
```
Cwejman, line 26: Given your updates in onChange() for Data form's input it should never be an "empty" string. Also I think that transform is not for validation but for parsing but perhaps I'm wrong. What is your idea behind it?

web/src/data/customer/queries.ts
```diff
@@ -9,11 +9,13 @@ import { differenceInYears, parse } from 'date-fns';
 
 //
 
-export const calculateAge = (persIdNo: string) =>
-  differenceInYears(
+export const calculateAge = (persIdNo?: string | null) => {
+  if (!persIdNo) return undefined;
```
Cwejman, line 13: In this case, a block function is unnecessary as it is an arrow function and could return a ternary. Ia bit of a nitty gritty detail but we have written in our guidelines that we prefer ternaries over if statements where it does not become too complicated.

web/src/data/employee/form.ts
```diff
@@ -13,5 +13,12 @@ export const employeeZ = z.object({
   email: reqString('Email').email({
     message: 'The email is not a valid email address',
   }),
-  personalIdentityNumber: swedishPIN,
+  personalIdentityNumber: z
+    .string()
+    .nullable()
+    .transform((val) =>
+      typeof val === 'string' && val.trim() === '' ? undefined : val,
+    )
```
Cwejman, line 21: Here though I suppose I understand the usage of transform as it is piped after into swedishPIN. ought again I think that it is unnecessary? As the onChange will not allow any "empty" strings

Cwejman (changes requested):
A few comments, nothing critical. have not run it locally though, is to be tested by someone else in the team.

Madasa-cell (commented):
Did some tests here are some critiques:

1. Maybe have a "error message" show itself for fields that have to be filled? For phone number and email. They do have it in edit but not create.
2. Personal ID placeholder could be updated. It currently suggests a YYMMDD-XXXX setup but will then demand a YYYYMMDD-XXXX personal identity nr.
3. Phone number can filled with nonsense that is most definitely not a phone number, maybe there is a reason for this...
    3b. In the future we could create a whitelist for acceptable symbols like numbers and '+'.
4. Update personal identity nr. in data.sql, you cannot do an edit for example "Christopher Lo-Martire" because his id does not follow the YYYYMMDD-XXXX format.
5. If you delete an employee that is used elsewhere and then go to said elsewhere such as Inventory, Tickets or Customers, then that page won't render. But this will be handled in an other issue apparently.

Otherwise it looks good.

daniel-svendsen:
looking good

Daviduber69 (approved):
All CRUD ops works, no console errors, every little thing, is gonna be allright

web/src/data/employee/form.ts
```diff
@@ -13,5 +13,5 @@ export const employeeZ = z.object({
   email: reqString('Email').email({
     message: 'The email is not a valid email address',
   }),
-  personalIdentityNumber: swedishPIN,
+  personalIdentityNumber: z.string().nullable().optional().pipe(swedishPIN),
```
Cwejman, line 16: why add string optional nullable when it is already built into swedishPIN?

Cwejman (changes requested):
One minor comment otherwise all good

web/src/data/customer/queries.ts
Cwejman: FYI here you could of written peerIdNo && calulcateAge( instead of the ternary but I'll mark as resolved

Cwejman (approved):

PR #177, opened 2025-03-04
feat(web): image delete endpoint

daniel-svendsen:
- now possible to delete image from S3 bucket and from object (also deletes it from S3)
- adds endpoint for deletion of imageURL
- adds querys for vehicle, general content and shop
- changes name from imageUrl to imageURL across several files

### DIFF
--- a/main/src/main/java/se/hjulverkstan/main/controller/ImageController.java
+++ b/main/src/main/java/se/hjulverkstan/main/controller/ImageController.java
@@ -6,6 +6,7 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.multipart.MultipartFile;
 import se.hjulverkstan.main.dto.ImageUploadResponse;
+import se.hjulverkstan.main.service.ImageService;
 import se.hjulverkstan.main.service.S3Service;
 
 @RestController
@@ -13,13 +14,23 @@ import se.hjulverkstan.main.service.S3Service;
 public class ImageController {
 
     private final S3Service s3Service;
+    private final ImageService imageService;
 
-    public ImageController(S3Service s3service) {
-        this.s3Service = s3service;
+    public ImageController(S3Service s3Service, ImageService imageService) {
+        this.s3Service = s3Service;
+        this.imageService = imageService;
     }
 
     @PostMapping(value = "/upload", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
     public ResponseEntity<ImageUploadResponse> uploadImage(@RequestParam("file") MultipartFile file) {
         return new ResponseEntity<>(s3Service.uploadFile(file), HttpStatus.OK);
+    }
+
+    @DeleteMapping("/delete")
+    public HttpStatus deleteImage(@RequestParam("imageURL") String imageURL) {
+        String fileKey = s3Service.extractKeyFromURL(imageURL);
+        s3Service.deleteFileByKey(fileKey);
+        imageService.deleteSpecificS3URLFromAllEntities(fileKey);
+        return HttpStatus.OK;
     }
 }
--- a/main/src/main/java/se/hjulverkstan/main/repository/ImageRepositoryImpl.java
+++ b/main/src/main/java/se/hjulverkstan/main/repository/ImageRepositoryImpl.java
@@ -1,0 +1,38 @@
+package se.hjulverkstan.main.repository;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.PersistenceContext;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.List;
+
+import java.util.stream.Collectors;
+
+@Repository
+@Transactional
+public class ImageRepositoryImpl implements ImageRepository {
+
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    private static final List<String> imageEntities = List.of("vehicle", "Shop", "General_Content");
+
+    @Override
+    public void deleteSpecificS3URLFromAllEntities(String fileKey) {
+        for (String entity : imageEntities) {
+            entityManager.createNativeQuery(
+                            "UPDATE " + entity + " SET imageURL = NULL WHERE imageURL LIKE CONCAT('%', :fileKey)")
+                    .setParameter("fileKey", fileKey)
+                    .executeUpdate();
+        }
+    }
+
+    @Override
+    public List<String> getAllUsedS3URLs() {
+        String unionQuery = imageEntities.stream()
+                .map(entity -> "SELECT imageURL FROM " + entity + " WHERE imageURL IS NOT NULL")
+                .collect(Collectors.joining(" UNION "));
+        return entityManager.createNativeQuery(unionQuery).getResultList();
+    }
+}
--- a/main/src/main/java/se/hjulverkstan/main/service/ImageServiceImpl.java
+++ b/main/src/main/java/se/hjulverkstan/main/service/ImageServiceImpl.java
@@ -1,0 +1,26 @@
+package se.hjulverkstan.main.service;
+
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+import se.hjulverkstan.Exceptions.InternalServerErrorException;
+import se.hjulverkstan.main.repository.ImageRepository;
+
+@Service
+@Transactional
+public class ImageServiceImpl implements ImageService {
+
+    private final ImageRepository imageRepository;
+
+    public ImageServiceImpl(ImageRepository imageRepository) {
+        this.imageRepository = imageRepository;
+    }
+
+    @Override
+    public void deleteSpecificS3URLFromAllEntities(String fileKey) {
+        try {
+            imageRepository.deleteSpecificS3URLFromAllEntities(fileKey);
+        } catch (Exception e) {
+            throw new InternalServerErrorException("Failed to delete image URL from database: " + e.getMessage(), e);
+        }
+    }
+}
--- a/main/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
+++ b/main/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
@@ -3,6 +3,7 @@ package se.hjulverkstan.main.service;
 import com.amazonaws.SdkClientException;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.model.CannedAccessControlList;
+import com.amazonaws.services.s3.model.DeleteObjectRequest;
 import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.amazonaws.services.s3.model.PutObjectRequest;
 import org.apache.tika.Tika;
@@ -13,6 +14,7 @@ import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.multipart.MultipartFile;
 import se.hjulverkstan.Exceptions.*;
 import se.hjulverkstan.main.dto.ImageUploadResponse;
+import se.hjulverkstan.main.repository.ImageRepository;
 
 import java.io.BufferedInputStream;
 import java.io.IOException;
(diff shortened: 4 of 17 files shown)


main/src/main/java/se/hjulverkstan/main/repository/ImageRepositoryImpl.java
Cwejman, line 1: This looks more like a service, using multiple repos than a repo it self. Not sure I see the point in doing multiple repos when it could be one query?

main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
```diff
@@ -99,26 +103,30 @@ public VehicleDto getVehicleById(Long id) {
 
     @Override
     public EditVehicleDto editVehicle(Long id, EditVehicleDto editVehicle) {
-        Vehicle selectedVehicle = vehicleRepository.findById(id).orElseThrow(() -> new ElementNotFoundException(ELEMENT_VEHICLE));
+        Vehicle selectedVehicle = vehicleRepository.findById(id)
+                .orElseThrow(() -> new ElementNotFoundException(ELEMENT_VEHICLE));
+
+        if (editVehicle.getImageURL() == null && selectedVehicle.getImageURL() != null) {
+            String fileKey = s3Service.extractKeyFromUrl(selectedVehicle.getImageURL());
+            s3Service.deleteFileByKey(fileKey);
```
Cwejman, line 111: This is not how the flow was planned. Currently that is the fronteds responsibility. Here you need to talk together the whole team and discuss which way to go

Cwejman (changes requested):

Daviduber69 (commented):
I tried this branch on the image-component branch and deleting images seems to be working as intended.
The imageURL is deleted from the vehicle object when clicking Save in Edit Mode and also deleted from the S3 bucket.
Checked in the portal and in postman to verify.

There are alot of out-commented code in ImageRepositoryImpl though?

daniel-svendsen:
> I tried this branch on the image-component branch and deleting images seems to be working as intended. The imageURL is deleted from the vehicle object when clicking Save in Edit Mode and also deleted from the S3 bucket. Checked in the portal and in postman to verify.
> 
> There are alot of out-commented code in ImageRepositoryImpl though?

the outcommented code is for soon-to-be added features

Daviduber69 (approved):

main/src/main/java/se/hjulverkstan/main/controller/ImageController.java
```diff
@@ -5,21 +5,36 @@
...
 @RequestMapping("/v1/image")
 public class ImageController {
 
     private final S3Service s3Service;
+    private final ImageRepository imageRepository;
 
-    public ImageController(S3Service s3service) {
+    public ImageController(S3Service s3service, ImageRepository imageRepository) {
         this.s3Service = s3service;
+        this.imageRepository = imageRepository;
```
Cwejman, line 22: imageRepository is added but is not used?

main/src/main/java/se/hjulverkstan/main/dto/ImageDeleteResponse.java
```diff
@@ -0,0 +1,17 @@
+package se.hjulverkstan.main.dto;
+
+public class ImageDeleteResponse {
+    private String message;
+
+    public ImageDeleteResponse(String message) {
+        this.message = message;
+    }
```
Cwejman, line 8: Not sure about the use of message here, have had this discussion with jimmy – talk to him about it

main/src/main/java/se/hjulverkstan/main/dto/ImageDeleteResponse.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public String getMessage() {
+        return message;
+    }
+
+    public void setMessage(String message) {
```
Cwejman, line 14: Regardless of comment above, here you could of used Lombok instead.

main/src/main/java/se/hjulverkstan/main/repository/ImageRepositoryImpl.java
Cwejman, line 1: Perhaps I'm blind but I cannot se the usage of the delete entities anywhere, as previously mentioned.

But what needs to be done here is to handle failure in a good way, and using @transaction would be best. The repository would therefore need to be used by a service that can apply the transaction.

main/src/main/java/se/hjulverkstan/main/repository/ImageRepositoryImpl.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+@Repository
+@Transactional
+public class ImageRepositoryImpl implements ImageRepository {
```
Cwejman, line 15: Also I think this repo could be improved quite a lot.

First of, creating sepparate native querys is not a problem I think as long as they are wrapped in a transaction as mentioned previously.

But since both methods here iterate over the same list of entities perhaps the repo could a a field on it

```java
private final List imageEntities = List.of("vehicle", "Shop", "General_Content");
```

I think using separate queries on the get method though, with null checks after might be a bit unessesary. With a field for the entities that method could be simpler like this:

```java
public List<String> getAllUsedS3URLs() {
    String sql = imageEntities.stream()
        .map(table -> "SELECT imageURL FROM " + table + " WHERE imageURL IS NOT NULL")
        .collect(Collectors.joining(" UNION "));

    return entityManager.createNativeQuery(sql).getResultList();
}
```

and the delete could be in a for loop instead:

```java
for (String entity : imageEntities) {
        entityManager.createNativeQuery(
            "UPDATE " + entity + " SET imageURL = NULL WHERE imageURL LIKE CONCAT('%', :fileKey)")
            .setParameter("fileKey", fileKey)
            .executeUpdate();
    }
```

main/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
```diff
@@ -102,12 +106,24 @@ private String generateUniqueFileName(MultipartFile file) {
...
+    public void deleteFileByKey(String key) {
+        try {
+            DeleteObjectRequest deleteRequest = new DeleteObjectRequest(bucketName, key);
+            amazonS3.deleteObject(deleteRequest);
+            System.out.println("Deleted file from S3 with key: " + key);
+        } catch (Exception e) {
+            System.err.println("Error deleting file from S3: " + e.getMessage());
+            throw e;
+        }
+        imageRepository.deleteSpecificS3URLFromAllEntities(key);
```
Cwejman, line 127: Oh, now I've found the usage of image repo.

This I think is bad practice, the s3 service should be a decoupled service only to do with s3, and image tweaks to be separate from that. So that the controller can orchestrate the separate concerns instead.

This together with the previous comment about adding an image service that wraps the repo in. transactions as well

Cwejman (changes requested):

main/src/main/java/se/hjulverkstan/main/controller/ImageController.java
```diff
@@ -5,21 +5,37 @@
...
+    public ResponseEntity<ImageDeleteResponse> deleteImage(@RequestParam("imageURL") String imageURL) {
+        try {
+            String fileKey = s3Service.extractKeyFromURL(imageURL);
+            s3Service.deleteFileByKey(fileKey);
+            imageService.deleteSpecificS3URLFromAllEntities(fileKey);
+            return new ResponseEntity<>(new ImageDeleteResponse("Image deleted successfully."), HttpStatus.OK);
+        } catch (Exception e) {
+            return new ResponseEntity<>(new ImageDeleteResponse("Error: " + e.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
```
jimmsi, line 40: The exception handling differs somewhat from ImageUpload. To follow the code pattern from ImageUpload, the try/catch block should be removed from the controller. In general for all files in this commit, try to work on converting exceptions into custom exceptions that can be handled globally by the ExceptionsController.

main/src/main/java/se/hjulverkstan/main/service/ImageServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public ImageServiceImpl(ImageRepository imageRepository) {
+        this.imageRepository = imageRepository;
+    }
+
+    @Override
+    public void deleteSpecificS3URLFromAllEntities(String fileKey) {
+        imageRepository.deleteSpecificS3URLFromAllEntities(fileKey);
+    }
+}
```
jimmsi, line 21: Perhaps the database interaction also needs exception handling?

main/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
```diff
@@ -102,12 +104,23 @@ private String generateUniqueFileName(MultipartFile file) {
     }
 
     @Override
-    public void deleteFileByKey(String key) {
-
+    public String extractKeyFromURL(String URL) {
+        String keyWithParams = URL.substring(URL.lastIndexOf("/") + 1);
+        int queryIndex = keyWithParams.indexOf("?");
+        String fileKey = queryIndex != -1 ? keyWithParams.substring(0, queryIndex) : keyWithParams;
+        System.out.println("Extracted file key: " + fileKey + " from URL: " + URL);
```
jimmsi, line 111: You use System.out.println (and also System.err) in many places to print what is happening (could be skipped?) or if something goes wrong. In ImageUpload, this is not used at all. However, perhaps we need to implement proper logging in the backend? This could be an open question, but if so, there are better alternatives, such as logger.error.

main/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
```diff
@@ -102,12 +104,23 @@ private String generateUniqueFileName(MultipartFile file) {
...
+    public void deleteFileByKey(String key) {
+        try {
+            DeleteObjectRequest deleteRequest = new DeleteObjectRequest(bucketName, key);
+            amazonS3.deleteObject(deleteRequest);
+            System.out.println("Deleted file from S3 with key: " + key);
+        } catch (Exception e) {
+            System.err.println("Error deleting file from S3: " + e.getMessage());
+            throw e;
+        }
     }
```
jimmsi, line 125: For the delete operation to S3, you could, for example, create a new S3DeleteException

jimmsi (commented):
Some comments mainly about dealing with exceptions.

main/src/main/java/se/hjulverkstan/main/controller/ImageController.java
jimmsi: Or use custom exceptions we already have defined

main/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
daniel-svendsen: was used during development, forgot to delete

main/src/main/java/se/hjulverkstan/main/controller/ImageController.java
```diff
@@ -5,21 +5,33 @@
...
         return new ResponseEntity<>(s3Service.uploadFile(file), HttpStatus.OK);
     }
+
+    @DeleteMapping("/delete")
+    public ResponseEntity<ImageDeleteResponse> deleteImage(@RequestParam("imageURL") String imageURL) {
+        String fileKey = s3Service.extractKeyFromURL(imageURL);
+        s3Service.deleteFileByKey(fileKey);
+        imageService.deleteSpecificS3URLFromAllEntities(fileKey);
+        return new ResponseEntity<>(new ImageDeleteResponse(imageURL,"Image deleted successfully."), HttpStatus.OK);
+    }
```
jimmsi, line 36: I'm a bit unsure whether the response object should include a message, as Jona previously pointed out and as we discussed. While the delete endpoint doesn’t have something concrete to return in the same way that the upload endpoint returns a URL, looking at other delete endpoints in the backend, we see that they don’t return a message to the client upon successful deletion. I feel that including a message in the response object might be unnecessary. Messages indicating that something went wrong can be handled by exception handling, which already communicates any errors, and if the operation succeeds, the response status code signals that everything went well. Perhaps others have different opinions on this?

main/src/main/java/se/hjulverkstan/main/dto/ImageDeleteResponse.java
Cwejman, line 1: This file is still here

Cwejman (changes requested):
Looks good.

Left one comment with a file that is no longer used but still committed.

Also the PR includes three commits that probably should be squashed as one.

Cwejman (changes requested):
Lastly just one comment

feat(web): image delete endpoint

should be feat(api): Image delete endpoint

as it is a backend change

Cwejman (approved):

PR #178, opened 2025-03-04
fix(web): cannot create customer owned vehicles

daniel-svendsen:
status and regtag was marked .nullable()
now is .optional()


Cwejman:
Perhaps this is resolved by #200 ?

Issue #179, opened 2025-03-04
Jules Baumman Code-test feedback issues


jules0606:
Backend issue : 

   One of the first thing that striked me on the backend side was how many differents vehicle and ticket classes were presents. It took time for me to analyze it and I do understand it is due to the way the team chose to develop these classes in the logic of using inheritance and also using hibernate mapping on top of that with  the InheritanceType.SINGLE_TABLE strategy. But I really feel like this is a bit of over-engineering implementation due to the will of the team trying to create a scalable solution. For example, It would also make the case where you have to add or change types of tickets in the future take more time to develop, and with more risks of bugs, because of the way the classes are dependent and similar. 
  In the current solution, the classes TicketDonate and TicketReceive both extends Tickets and have the exact same implementation of isValidTicketStatusTransition(). If these classes are so similar and are just copy of the ticket Class, why should they exist?
   In the end I feel like it created a lot of complexity during the implementation of TicketServiceImpl.java Class and I also think it will be harder to maintain in the long run than a solution without inheritance. From my point of view, the solution should have stayed more simple with the type of the ticket being part of the Ticket Class and for example the endDate also being part of it but only being used depending of the type of the ticket.

Frontend issue : 

  I don't have react experience so I couldn't really see something in the frontend codebase, but I would like to use the experience I have as an angular developer to discuss the choices over the ui/ux. I really appreciate the team's efforts to make the main data-table good to look at and with a lot of filter options and features that are pertinent for these kind of applications. But I also think that there are a lot of informations already displayed on the screen, it is a lot for the user and they might be lost on the first screen. 
From my perspective, for example, I would have preferred if the "create vehicle" on this screenshot for instance was part of either another screen or a dialog-type screen to make the form bigger and easier to read/interact with :
 https://github.com/Hjulverkstan/hjulverkstan/blob/code-test/.screenshots/250218-PortalShopInventory4.png 
I also think the 'main' screen should contain only the most important informations for the user and the rests of the informations available should in a different detail screen/dialog.

PR #180, opened 2025-03-04
issue #179 : Jules Baumman Code-test feedback issues pull request

jules0606:
Regarding the issue #179 I submitted earlier regarding the backend and the different Ticket classes, I implemented the start of my view on how I would simplify all of these classes and the TicketServiceImpl.java subsequently with all these changes.
-First, I deleted all the "instance" type classes.
-I implemented a util class for the validator to make sure the data received front the frontend will still be valid.
-I then modified a lot of the code from the service and I think it makes it a lot more readable and easy to maintain.


Cwejman:
Hi again Jules,

This PR was part of a code test where the submission was supposed to be private until the interview process is over.

That the PR ended up here is fine though no worries!

The spring application is indeed "in need" of many improvements and the idea to simplify and move away from an inheritance pattern where perhaps not needed is appreciated.

I think though, that the moving of validation logic to a separate util is a bit bad practice when it can reside in the model it self, aligning with your original intent to defragment and unify logic.

If you would like to continue to refine this PR that is very welcome but you may of course close it as well if you'd like to let it go.

Please report back so that we can keep the PRs from going stale.

Thanks,
Jona

jules0606:
Hi Jona,

I am sorry about the misunderstanding about the PR, It does make more sense that my work was supposed to be private during the interview process and I should have been more careful when I submitted my work.

About The solution itself, I thought and implemented what I believe is a better DTO validation logic, meanwhile the validations annotations in the models will be used when the controller methods are called.
I think my changes keep in line with my original logic of simplifying and removing the tickets subclasses, the main outcomes will be:

Maintainability: Validation logic is now centralized and reusable.
Clarity: Error messages are standardized and user-friendly.
Scalability: New ticket types can be added with minimal code changes.

If you still want me to close the PR for now until the interview process is over then I will do it without problem, I could keep this work in a private repository meanwhile, please let me know how you want to handle this.

Thank you, Best Regards,
Jules

Cwejman:
> Hi Jona,
> 
> I am sorry about the misunderstanding about the PR, It does make more sense that my work was supposed to be private during the interview process and I should have been more careful when I submitted my work.
> 
> About The solution itself, I thought and implemented what I believe is a better DTO validation logic, meanwhile the validations annotations in the models will be used when the controller methods are called. I think my changes keep in line with my original logic of simplifying and removing the tickets subclasses, the main outcomes will be:
> 
> Maintainability: Validation logic is now centralized and reusable. Clarity: Error messages are standardized and user-friendly. Scalability: New ticket types can be added with minimal code changes.
> 
> If you still want me to close the PR for now until the interview process is over then I will do it without problem, I could keep this work in a private repository meanwhile, please let me know how you want to handle this.
> 
> Thank you, Best Regards, Jules

No worries, but now that the PR is here we either merge it to the project with our values aligned or leave it be and close it. And I think that, while the intent of the PR, removing the inheritance, is good and the project would benefit from it I think that the TicketUtil could be removed in favour of validation on the ticket model it self. Given that the ticket model is used in the constructor of DTO's as well, I see no need to decouple the validation away from the optimal place for it. But perhaps I'm missing something or we have a different interpretation of best practices?

To clarify, if you want to work on the PR and contribute with it to the project we will have to polish it to a state where we both feel it is ready. Appart from my feedback it would also be sound to invite one of the interns currently maintaining the project for a round of peer review as well, but lets start with the point at hand.

Cwejman:
Apologies for the lengthy thread, but as I mentioned when we meet previously this week, I have much left to learn with Java and Spring. I'm keen integrating my existing values (much of which I've condensed into [GUIDELINES.md](https://github.com/Hjulverkstan/hjulverkstan/blob/main/GUIDELINES.md)) into Java OOP and getting the best of the two.

While the idea above with the bean wrapper satisfies a part of my self, thinking more about I see that it is most likely a bit nonstandard, typeless and feels a bit "hacky".

What do you think about unifying into just one dto instead? Our application only uses PUT so far not PATCH so the dtos are very similar between create, edit and the response.

But I know that having separate dtos is usual in spring applications and so if your approach is the standard approach then perhaps there is not much to critique. With this project though, compared to billed projects were we have to be effective and cannot go back and refactor as much without billing the client more, here we have the time refine our work more – in my opinion a great platform for learning and of course gratifying to a perfectionist.

Issue #30, opened 2024-06-07
Mobile inventory page


Cwejman:
When opening portal on mobile, show a custom mobile adapted image uploader page. It should have a drop down for the vehicle and the possibility to upload an image for that vehicle. Maybe also seeing the existing image for the selected vehicle?

## Frontend

To keep all this as simple as possible and so that we don't spend our time squeezing in mobile support for our existing pages. If on mobile the page will be conditionally rendered to only render the custom mobile image uploader page. This is done in root/Portal/index.tsx. Which means we are creating a new page exclusively for mobile so go to /web/src/root/portal and create a new folder with a fitting name. Similar files needs to be created as those in the adjacent directories.

To start with the mobile page should only include, vertically displayed cards for each vehicle. Sorted by latest createdAt. Every card should have background image of either the corresponding vehicle or a placeholder if no image has been assigned to it. To keep it simple the regTag should be rendered below the image card, not on top or as an overlay. When you press the image (or regTag) you should be redirected to the image upload/change page, and on success you trigger the vehicleEditMutation to update the vehicle's image URL.

------------

The component to upload (and delete) images is done in task #29 so the comment about "It should have a drop down for the vehicle and the possibility to upload an image for that vehicle." is something I interpret as "include the functionality of the work done from this task".

The Search and Filter Function has been refactored out into its own task (#145) as it was never asked for in this task but has potential usability for users of the app. This was only done to make this Task get done faster, however it has grown larger so maybe this has become useless?

## Backend
Nothing, ImageURL should already have been implemented.

## Design Spec

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/b34d4eb8-2d2d-4323-9bf2-6bf5d1321ecf" />

Cwejman:
Perhaps the best way is to use tailwind media query with toggling display: none so that we do not have a seperate path for this mobile page using redirects but just, conditionally using css always render the mobile page when on mobile.

Or a useEFfect that redirects to this url.

Needs to be thought through.

PR #181, opened 2025-03-06
feat(web): Mobile inventory page

Madasa-cell:
If on mobile a special page will be conditionally rendered.

This is for ease of uploading images.

Solves issue #30

-----------

There are some thing I'll need to rebase from other branches but those aren't done yet. So while the functionality of the page is not done, the design is, feel free to criticize it and how I'am currently connected to these yet to be rebased branches.

### DIFF
--- a/web/src/root/Portal/PortalContent.tsx
+++ b/web/src/root/Portal/PortalContent.tsx
@@ -8,7 +8,7 @@ export default function PortalContent({ children }: PortalContentProps) {
   return (
     <div
       className="bg-background css-glow-shadow mb-4 flex min-h-0 flex-1
-        overflow-hidden rounded-md border"
+        overflow-hidden border sm:rounded-md"
     >
       {children}
     </div>
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -39,6 +39,7 @@
         "react-router-dom": "^6.22.1",
         "tailwind-merge": "^2.2.1",
         "tailwindcss-animate": "^1.0.7",
+        "vaul": "^1.1.2",
         "zod": "^3.22.4"
       },
       "devDependencies": {
@@ -9005,6 +9006,19 @@
       "license": "MIT",
       "engines": {
         "node": ">= 0.8"
+      }
+    },
+    "node_modules/vaul": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/vaul/-/vaul-1.1.2.tgz",
+      "integrity": "sha512-ZFkClGpWyI2WUQjdLJ/BaGuV6AVQiJ3uELGk3OYtP+B6yCO7Cmn9vPFXVJkRaGkOJu3m8bQMgtyzNHixULceQA==",
+      "license": "MIT",
+      "dependencies": {
+        "@radix-ui/react-dialog": "^1.1.1"
+      },
+      "peerDependencies": {
+        "react": "^16.8 || ^17.0 || ^18.0 || ^19.0.0 || ^19.0.0-rc",
+        "react-dom": "^16.8 || ^17.0 || ^18.0 || ^19.0.0 || ^19.0.0-rc"
       }
     },
     "node_modules/vite": {
--- a/web/package.json
+++ b/web/package.json
@@ -48,6 +48,7 @@
     "react-router-dom": "^6.22.1",
     "tailwind-merge": "^2.2.1",
     "tailwindcss-animate": "^1.0.7",
+    "vaul": "^1.1.2",
     "zod": "^3.22.4"
   },
   "devDependencies": {
--- a/web/src/components/ConfirmDeleteDialog.tsx
+++ b/web/src/components/ConfirmDeleteDialog.tsx
@@ -13,7 +13,7 @@ interface confirmDeleteDialogProps {
   onDelete: () => void;
   onCancel?: () => void;
   entity: string;
-  entityId: string;
+  entityId?: string;
 }
 
 export default function ConfirmDeleteDialog({
@@ -23,14 +23,14 @@ export default function ConfirmDeleteDialog({
   entityId,
 }: confirmDeleteDialogProps) {
   const item = entity.toLowerCase();
-  const id = entityId.toLowerCase();
+  const id = entityId?.toLowerCase();
 
   return (
     <DialogContent className="sm:max-w-[425px]">
       <DialogHeader>
         <DialogTitle>{`Delete ${item}`}</DialogTitle>
         <DialogDescription>
-          {`You are trying to delete a ${item} with id ${id}.`}
+          {`You are trying to delete a ${item} ${id ? `with id ${id}.` : ''}`}
         </DialogDescription>
       </DialogHeader>
       <DialogFooter>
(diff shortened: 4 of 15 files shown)


Daviduber69 (commented):
Looks visually nice!
As of now (which Adam already knows) you cant Edit to upload new images. Which is only about changing the createCreateImage POST req in image/api (and also its use in mutation)

And also the faulty image URL isn't being handled so 
image
it looks like this now, maybe have a placeholder fallback message/image for that.

The no-image-image looks good tho!

------------------------

edit:
the picture scaling look really good!

maybe make the upload+delete buttons smaller,?
enable all vehicles to have images,
look into handleImageUpload and not have dragEvents (?),
Maybe add a Loading indicator while upload is pending ?


image

Also here you can use variant instead of coding your own design to match the rest of the app's design.

web/src/components/ImageBox.tsx
```diff
@@ -1,35 +1,202 @@
...
+    if (!file || uploadImageM.isPending) return;
+
+    uploadImageM
+      .mutateAsync({ file })
+      .then((imageURL) => {
+        // Update local state with the new image URL.
+        onChange(imageURL);
+        // Update the backend.
+        return onBackendUpdate({ [dataKey]: imageURL });
+      })
```
jimmsi, line 79: Here, the image upload logic immediately triggers an update to the backend (onBackendUpdate) right after the upload completes. This differs from how updates are handled in other parts of the code, such as PortalForm.tsx, where changes are first applied to the UI and then saved separately to the backend. Would it make sense to follow a similar pattern here? Updating the UI first and then triggering the backend update separately? Could improve consistency, prevent potential race conditions, and provide better user feedback—especially in cases where the backend update fails.

web/src/components/DataForm/DataFormMobileImage.tsx
```diff
@@ -0,0 +1,222 @@
...
+        setShowOptions(false);
+      }
+    };
+    if (showOptions) {
+      document.addEventListener('mousedown', handleClickOutside);
+    }
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, [showOptions]);
```
jimmsi, line 159: The mousedown event listener is currently added and removed every time showOptions changes. This increases complexity and performance costs, especially if the component re-renders frequently. Other parts of the codebase, such as PortalForm.tsx, use more stable triggers (e.g., screen width) or direct UI interactions rather than global event listeners. Would it be better to keep the event listener stable by using a useRef, so it only initializes once and doesn’t depend on state changes?

Cwejman:
When stating to review it seems there is a load of commits here that should not be part of the PR.

It has to be tidied up before I can review.

One method to resolve it in intellij to duplicate this branch, then reset it to main so that it is identical to main. After that you could cherry pick your only commit from you duplicate branch. Right click and cherry pick in intellij.

web/src/root/Portal/PortalForm.tsx
```diff
@@ -94,7 +94,9 @@ export default function PortalForm({
   }[mode];
 
   return (
+    // Contains the PortalForm
```
Cwejman, line 97: Remove this or make separate commit to reduce footprint

web/src/root/Portal/PortalContent.tsx
```diff
@@ -8,7 +8,7 @@ export default function PortalContent({ children }: PortalContentProps) {
   return (
     <div
       className="bg-background css-glow-shadow mb-4 flex min-h-0 flex-1
-        overflow-hidden rounded-md border"
+        overflow-hidden border sm:rounded"
```
Cwejman, line 11: Same here, reducing footprint. These changes are typically a chore.

Cwejman (changes requested):
Notes from discussions in person.

Lets simplify how this works.

Some features have exceeded the scope of the issue:

- Using datable (we keep this as it will be valuable the day we want to introduce filters)
- Adding upload and delete buttons that appear when clicking a card (this is also valuable – what if you upload a wrong image and then cant remove it, upload to the wrong bike for instance?)

In the spirit of simplifying let's make the bike card dumb instead (and rename from box to card). Given it being dumb it can even be in the same file if we want to as it will not be too long, or otherwise in the same dir as the global components dir is not correct.

NOTE: In a few days there will be a ready made card component that is used on the public website so that can replace – lets just do something simple now that will work.

The image card component will not ahve error or loading logic, it is just a dumb card with an on click handler.

To keep the idea of multiple actions on card we can use the drawer from shadcn

Install that component.

Go thorugh the drawer file and align it so it looks like other components import from shadcn in term of naming and export. Check table it has a coherent style for shadcn component – looking at naming and exports etc.

Wrap the image card in the drawer in the index.tx file and in the drawer content add a drawer close (cancel), delete action and upload action.

For delete lets use a confirm dialog like we do in other parts of the app.

For upload lets do the simpler recepie

```typescript
const fileRef = useRef<HTMLInputElement>(null);

return (
  <>
    <input
      ref={fileRef}
      type="file"
      id="file-upload"
      className="hidden"
      onChange={handleUpload}
    />
    <Button onClick={() => fileRef.current?.click()}>Upload</Button>
  </>
);
```

To programatically trigger image upload where handle upload. To keep the component simple and use the built in error and pending mechanics of react query we'll experiment and try a new approach of combining multiple api functions in one mutation hook function:

```typescript
export const useUploadAndSetVehicleImage = () => {
  const { mutationFn: uploadFn } = api.createUploadImage();
  const { mutationFn: editFn } = api.createEditVehicle();

  useMutation({
    mutationFn: (file: string, vehicle: Vehicle) => uploadFn({ file }).then(imageURL => editFn({ ...vehicle, imageURL })
    onSuccess: ({ id }) => invalidateQueries([ api.createGetVehicles().queryKey  ]),
  });
```

And, when uploading and setting, aka when the isPending prop is true for the mutation hook disable all buttons in drwawer, can conditionally render an disabled icon button instead that has the spinner as the icon so that the upload button has a spinner on it.

web/src/root/Portal/PortalContent.tsx
Madasa-cell: I don't agree with this being a chore. It affects how the mobile page looks, without it the page doesn't look as good. This was the only place I could find to fix it.

Madasa-cell:
Here is an alternate version to ImageCard.tsx. Created because the images with dead links look bad but kept outside of a commit because the component is supposed to be kept dumb and not be capable of onError logic which this one is. Also since being able to see if a link is dead might be a good feature it now displays them as red vehicle icons compared to the usual grey icons.

```typescript
import React, { ComponentType, forwardRef, useEffect, useState } from 'react';

export interface ImageCardProps {
  imageURL: string;
  onClick: () => void;
  icon?: ComponentType<{ className?: string }>;
}

export const ImageCard = forwardRef<HTMLDivElement, ImageCardProps>(
  ({ imageURL, onClick, icon: Icon }, ref) => {
    const [isBroken, setIsBroken] = useState(false);

    useEffect(() => {
      setIsBroken(false);
    }, [imageURL]);

    const isImageMissing = !imageURL || imageURL === 'OK';
    const isValidImageURL = !isImageMissing && !isBroken;

    return (
      <div
        onClick={onClick}
        ref={ref}
        className="relative flex h-full w-full items-center justify-center
          overflow-hidden rounded-xl border bg-gray-100 hover:cursor-pointer"
      >
        {isValidImageURL ? (
          <img
            src={imageURL}
            alt=""
            onError={() => setIsBroken(true)}
            className="relative aspect-video w-full object-cover"
          />
        ) : Icon ? (
          <Icon className={`h-16 w-16 ${
            isBroken ? 'text-red-400' : 'text-gray-400'
          }`} />
        ) : (
          <div className="text-gray-400">No Image</div>
        )}
      </div>
    );
  },
);
```

Cwejman (approved):

PR #183, opened 2025-03-10
feat(web): add item warnings for vehicles and tickets

jimmsi:
Implement warnings that display icons and tooltips before IDs. Extend aggregated types with a warnings array and integrate a warning module (DUE_PICKUP, DUE_RETURN, ORPHAN). Extract shared logic into a helper.

### DIFF
--- a/web/src/root/Portal/PortalShopInventory/useColumns.tsx
+++ b/web/src/root/Portal/PortalShopInventory/useColumns.tsx
@@ -8,6 +8,7 @@ import { useMemo } from 'react';
 import { useLocationsAsEnumsQ } from '@data/location/queries';
 import { TicketBadges } from '../PortalShopTickets/useColumns';
 import { format } from 'date-fns';
+import WarningBadge from '@components/WarningBadge';
 
 //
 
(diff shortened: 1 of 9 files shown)


Daviduber69 (commented):
Everything seems to work fine!

Maybe instead of duplicating code in ticket/vehicle's useColumns, extract out WarningIcon and Tooltip
to its own component or something, and then use those components in useColumns?

web/src/components/Error.tsx
Cwejman, line 1: Why remove line breaks in this file=

web/src/root/Portal/PortalShopInventory/useColumns.tsx
Cwejman, line 1: I think the warnings icons part is a bit unfortunate, I see the point of making the warning enums own the icon but the solution is a bit complicated.

Also why does the warning icon need styling when the other icons dont have it?

Perhaps the icon matching could be simplified:

```typescript
icon: warnings?.length ? findWarning(warnings[0]) : baseBadge.icon
```


FYI, instead of hasWarnings one way to simplify checking if an array has element is to just `warnings.length` directly. As 0 is falsy and not 0 truthy.

Also if selected I think the variant should change to contrast still not stay destructive.

web/src/root/Portal/PortalShopTickets/useColumns.tsx
Cwejman, line 1: Same feedback applies here

Cwejman (changes requested):

web/src/components/Error.tsx
Cwejman: Still a question

Cwejman:
> Everything seems to work fine!
> 
> Maybe instead of duplicating code in ticket/vehicle's useColumns, extract out WarningIcon and Tooltip to its own component or something, and then use those components in useColumns?

I think this is a nice idea! I recommend to do it

Cwejman (approved):

Issue #184, opened 2025-03-10
Rent ticket still available to pick from Inventory Action menu when "Create Ticket"


Daviduber69:
**#186  this bug issue is linked here and there were more bugs that arose when I fixed this one**

We fixed a Bug where you could pick Customer Owned Vehicles on Rental Tickets, which you no longer can

But in the Action Menu of inventory when selecting "Create Ticket" on a Customer Owned Vehicle you can still select Rent, (but still can't select any customer owned vehicles from the List)

![Image](https://github.com/user-attachments/assets/2d180675-4d8a-4158-a41f-491d1e67dc99)

![Image](https://github.com/user-attachments/assets/d77f8e8e-052c-4a49-a52e-ace38db42e90)

![Image](https://github.com/user-attachments/assets/440e5b6b-df4e-4312-a75f-b9c13fd45621)

So the Rental ticket option should be disabled when going from the Action Menu -> Create Ticket
on Customer Owned Vehicles.

### BE
since we aldready made validation and exception management for customerOwned and Rental Tickets in the previous Bug Fix regarding this, that should be enough for this issue aswell, 
but in 
### FE
Look into PortalShopTicket/index on this part.
![Image](https://github.com/user-attachments/assets/6ccf0623-0bb5-4759-b3e8-b1ae1931a58b)

PR #186, opened 2025-03-11
fix(web): Rental ticket available on Create Ticket even though it shouldnt be

Daviduber69:
#184 Fixed so that when using the action menu from Inventory and selecting Create Ticket, when you select a customerOwned Vehicle TicketType RENT no longer appears in the list of tickets on the Ticket Page, but still do for Hjulverkstan Owned vehicles.
 Same with Donate Tickets.  
 And also it's mandatory to select a vehicle when creating tickets.


daniel-svendsen (approved):
working as intended from what i can see

jimmsi (commented):
Good catch! However, when creating tickets (whether from tickets or the action menu in inventory), it's possible to select "Repair," then pick a customer-owned bike, and afterward change the ticket type to "Rent." Although the ticket ultimately can't be created ("failed to create..."), the option to switch types shouldn't be available once a customer-owned bike is selected.

Cwejman (approved):

Issue #165, opened 2025-02-21
Security Issues, cannot Create new Users due to RefreshToken early expiration


Madasa-cell:
# Description

This is a problem related to the security.

Personally I believe the change occurred when I in #155 removed the id field from users in data.sql. But it could have been something else. Clearly the problem is closely connected with Signup.

POST | http://localhost:8080/v1/user/signup

This is where the 400 error occurs. The lowest point of our code in the Stacktrace is createCreateUser in user/api.ts 

## How to Trigger the Bug

1. Go to Admin/Users
2. Open DevTools to check up Storage and see that there exists a refreshToken cookie.
3. Go to the Console or Network section in DevTools (personal testing showed that the cookie for some reason won't disappear if you don't, the errors still remains even if you stay in this section).
4. Start creating a couple of new users
5. After 30 seconds you get a "Failed to create new users" Error Toast
6. In Network you get a 400 error about Signup failing.
7. In Storage the refreshToken has disappeared.

### These are some quick fixes for while you are using the Portal.
1. Move out of the Admin Section to any other such section, and you get a refreshtoken.
2. Move out of the Users Page into either Locations or Employees, and you get a refresh token.
3. Move out of Create Mode, press on a User to see the edit button, and you get a refresh token.
4. Put a browser window over the window that the portal is open in, and you get a refresh token.

Its worth noting that even when you get a refreshtoken that it doesn't reappear in the Storage section of DevTools. Unless you specifically press the cookie drawer, but when you are there it won't disappear either after 30 seconds.

### Other similar anomalies on other pages and sections.

1. When creating a Location, waiting 30 seconds, it throws out a 401 error and then immediately followed by a successful refreshtoken, if you don't have DevTools open a user wouldn't notice it.
2. The same is true for Vehicle, Customer, Ticket and Location.
3. Couldn't test Employee due to reasons. it kept telling me to fill in the fields despite them already being filled in.

## Frontend

TODO

## Backend

TODO

PR #199, opened 2025-03-19
fix(web): Signup on permitAll

Daviduber69:
In WebSecurityConfig signUp endpoint was in the permitAll requestMatchers, with this removed the crash in Users when refreshtoken expired no longer occurs.
 
 [#165](https://github.com/Hjulverkstan/hjulverkstan/issues/165)

### DIFF
--- a/main/src/main/java/se/hjulverkstan/main/config/WebSecurityConfig.java
+++ b/main/src/main/java/se/hjulverkstan/main/config/WebSecurityConfig.java
@@ -74,7 +74,6 @@ public class WebSecurityConfig {
                     auth.requestMatchers("/v1/auth/login",
                                     "/v1/auth/refreshtoken",
                                     "/v1/auth/signout/*",
-                                    "/v1/user/signup",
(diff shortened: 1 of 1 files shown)


Cwejman (changes requested):
I think we can be more direct here. This resolves not being able to create users after token refresh.

What I usually say is: write the commit from the user perspective if possible rather then technically.

It should also be linked to the bug in question

Daviduber69:
> > I think we can be more direct here. This resolves not being able to create users after token refresh.
> > What I usually say is: write the commit from the user perspective if possible rather then technically.
> > It should also be linked to the bug in question
> 
> I'm curious though, does that mean that Adam adding the auditable fields to user was not needed to resolve the bug?
> 
> That leaves his PR no longer being a bug fix and should be thought of accordingly.

I could not get it to work when Auditable extension was removed.

Maybe someone else can check it out aswell @jimmsi  @daniel-svendsen

Cwejman (approved):

PR #203, opened 2025-03-21
Fix/crash when removing employee

Daviduber69:


Madasa-cell (approved):
## Potential Problems
1. When you hover over the Delete button the Tooltip doesn't look particularly good. Looking at other tooltips they feature a triangle at the bottom pointing at the "problem" making it look like a speech bubble, some cast a shadow, most have slightly rounded corners and there is some padding to the side to not make it feel so cramped. However this was never a problem this task set out to fix, I just think it should be fixed since it looks bad.

Here's a solution to problem 1. The problem also occurs in Inventory.

```typescript
<Tooltip.Content
                side="top"
                align="center"
                className="bg-primary relative rounded-lg px-3 py-2 text-white
                  shadow-lg"
              >
                <div
                  className="border-t-primary absolute -bottom-2 left-1/2 h-0
                    w-0 -translate-x-1/2 transform border-l-8 border-r-8
                    border-t-8 border-l-transparent border-r-transparent"
                ></div>
                Cannot delete employees with tickets.
              </Tooltip.Content>
```

## Conclusion
I think this looks good, sure I have that slight problem with it, but this PR was never supposed to solve my issue with the design of Tooltip.Content.

web/src/root/Portal/PortalShopTickets/useColumns.tsx
```diff
@@ -45,7 +45,7 @@ export function TicketBadges({ ticketIds }: { ticketIds: string[] }) {
 
       return {
         ...ticketEnum,
-        href: `${coreUrl}/ticketz/${ticketId}`,
+        href: `${coreUrl.replace(appSlug, '/shop')}/ticketz/${ticketId}`,
```
Cwejman, line 48: This looks a bit tacky, usePortalSlugs should give you the needed prefix, replacing does not look particularly clean

main/src/main/java/se/hjulverkstan/main/service/EmployeeServiceImpl.java
```diff
@@ -58,6 +59,7 @@ public EmployeeDto deleteEmployee(Long id) {
             List<Ticket> tickets = employee.getTickets();
             //TODO: how to handle removing employee here?
```
Cwejman, line 60: We've decided to disallow removing employees with tickets for now until we implement soft delete.

But the backend here now allows it. The backend should be aligned with the frontend

Cwejman (changes requested):

Cwejman (approved):

PR #206, opened 2025-03-24
fix(web): No error toast when failing to upload images.

Daviduber69:
#205 added a Error Toast in the Image Upload Function so that when an image fails to upload, an error toast stating that pops up.


Cwejman (approved):
No comments on the code, has anyone tested this locally?

Issue #147, opened 2025-02-07
S3 Orphan File Cleanup


jimmsi:
S3 Orphan File Cleanup:
Create a new directory in backend called /jobs, and a new class S3OrphanFileCleanup. 
It periodically scans the S3 bucket to identify files (using their unique file keys) that are not referenced by any objects (such as vehicles, events, etc.) in the database. These files, deemed “orphan files,” are then automatically deleted from S3.

Setup execution interval
Use Spring’s @ Scheduled annotation to run the cleanup process automatically at a defined interval (for example, every hour).
Configure the execution interval via a property in application.properties. 

Implementation
- Create method getAllUsedS3URLs in imageRepository. It should store all S3URLs attached to entities in a List.
- Create method getAllImageKeys in S3Service. It should store all imageKeys from s3bucket in a list.
- For each imageKey in s3bucket, check the List with S3URLs attached to entities for a matching URL. 
- If no matches are found, add the imageKey to another List, orphans.
- For each imageKeys in orphans, use S3Service (method: deleteFileByKey) to delete imageKeys from s3.  

Error Handling and Logging:

Implement robust error handling to capture and log any issues that occur during S3 operations or database updates. Ensure that failures in one part of the process do not break the overall cleanup operation.

Cwejman:
I propose to make a jobs directory, since other jobs in the future will probably be created as well

jimmsi:
Directory jobs - class: cleanupjob Schedule annotation - call getAllUseds3URLs 

image repository
- methods, getAllUseds3URLs, deleteAllImagesByURL

s3 service - getAllImageURLs, deleteFileByURL

Delete endpoint - deleteFileByURL, deleteAllImagesByURL

PR #210, opened 2025-03-27
feat(api): S3 Orphan File Cleaner

Madasa-cell:
New class is created to clean (delete) orphaned file (image) keys from the S3 bucket. Orphaned files are those that are not referenced by any objects.

Resolves: #147

### DIFF
--- a/api/src/main/java/se/hjulverkstan/main/MainApplication.java
+++ b/api/src/main/java/se/hjulverkstan/main/MainApplication.java
@@ -3,8 +3,10 @@ package se.hjulverkstan.main;
 import org.springframework.boot.SpringApplication;
 
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
 import se.hjulverkstan.main.config.DotenvApplicationContextInitializer;
 
+@EnableScheduling
 @SpringBootApplication
 public class MainApplication {
 	public static void main(String[] args) {
--- a/api/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
+++ b/api/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
@@ -2,10 +2,7 @@ package se.hjulverkstan.main.service;
 
 import com.amazonaws.SdkClientException;
 import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.model.CannedAccessControlList;
-import com.amazonaws.services.s3.model.DeleteObjectRequest;
-import com.amazonaws.services.s3.model.ObjectMetadata;
-import com.amazonaws.services.s3.model.PutObjectRequest;
+import com.amazonaws.services.s3.model.*;
 import org.apache.tika.Tika;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
@@ -14,19 +11,16 @@ import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.multipart.MultipartFile;
 import se.hjulverkstan.Exceptions.*;
 import se.hjulverkstan.main.dto.ImageUploadResponse;
-import se.hjulverkstan.main.repository.ImageRepository;
 
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.Arrays;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
+import java.util.stream.Collectors;
 
 @Service
 @Transactional
 public class S3ServiceImpl implements S3Service {
-
     private final AmazonS3 amazonS3;
     private static final List<String> ALLOWED_MIME_TYPES = Arrays.asList("image/jpeg", "image/png", "image/gif");
     private static final Tika TIKA = new Tika();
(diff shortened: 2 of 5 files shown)


main/src/main/java/se/hjulverkstan/main/jobs/S3OrphanFileCleanup.java
```diff
@@ -0,0 +1,53 @@
...
+        this.s3ServiceImpl = s3ServiceImpl;
+    }
+
+    @Scheduled(cron = "${cleanup.s3-orphan-files.cron}")
+    public void runCleanup() {
+        List<String> usedUrls = imageRepositoryImpl.getAllUsedS3URLs();
+        List<String> allImageKeys = s3ServiceImpl.getAllImageKeys();
+
+        // For each imageKey in the s3bucket, check the list with used s3URLs attached to entities for a matching URL.
+        Set<String> usedKeys = usedUrls.stream()
```
Cwejman, line 34: Important to be attentive to the existing solution. There is already a method from s3 service that extracts key from url, and seems to ahve more safety to it as well. Use it instead?

main/src/main/java/se/hjulverkstan/main/jobs/S3OrphanFileCleanup.java
```diff
@@ -0,0 +1,53 @@
...
+                        .map(url -> url.substring(url.lastIndexOf("/") + 1))
+                        .collect(Collectors.toSet());
+
+        // If no matches for usedKeys are found, the imageKey is added to the list orphanKeys.
+        List<String> orphanKeys = allImageKeys.stream()
+                        .filter(key -> !usedKeys.contains(key))
+                        .toList();
+
+        // For each imageKey that is orphaned, S3Service deletes those imageKeys from the s3bucket.
+        for (String key : orphanKeys) {
```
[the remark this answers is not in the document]
Cwejman: Without reading up on s3 documentation it sounds reasonable to always use batch?

We create a list here so lets invoke s3 delete with a list directly.

main/src/main/java/se/hjulverkstan/main/jobs/S3OrphanFileCleanup.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.scheduling.annotation.Scheduled;
+import se.hjulverkstan.main.repository.ImageRepositoryImpl;
+import se.hjulverkstan.main.service.S3ServiceImpl;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Configuration
+@EnableScheduling
```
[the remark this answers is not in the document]
Cwejman: Yes I have previously seen the @EnabledScheduling annotation on a general web config class.

I propose to resolve this before closing the PR

Cwejman (changes requested):
@jimmsi has added good comments – agree with them. I added one other tweak as well.

api/src/main/java/se/hjulverkstan/main/MainApplication.java
```diff
@@ -3,8 +3,10 @@
 import org.springframework.boot.SpringApplication;
 
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
 import se.hjulverkstan.main.config.DotenvApplicationContextInitializer;
 
+@EnableScheduling
 @SpringBootApplication
 public class MainApplication {
 	public static void main(String[] args) {
```
jimmsi, line 12: I suggest we move the @EnableScheduling annotation from MainApplication to a more appropriate configuration class. Maybe a new AppConfig or SchedulingConfig class.

api/src/main/java/se/hjulverkstan/main/service/S3ServiceImpl.java
```diff
@@ -120,4 +114,37 @@ public void deleteFileByKey(String key) {
...
+     */
+    @Override
+    public List<String> getAllImageKeys() {
+        // A request object that tells S3 to list objects (files) from a specified bucket.
+        ListObjectsV2Request request = new ListObjectsV2Request().withBucketName(bucketName);
+        // The response from S3 that contains a list of files stored in the bucket.
+        ListObjectsV2Result result = amazonS3.listObjectsV2(request);
+
+        // Retrieves all file details from the bucket. Converts it to a stream.
+        // Extracts just the file key. Converts the stream back to a List<String>.
```
jimmsi, line 144: Some of the comments explain what the code is doing in a way that's already obvious from the code itself (e.g., describing standard Java Streams or basic S3 operations). Consider removing these and if neccesary add comments that explain why a piece of logic exists or any edge cases it's handling. This helps keep the code clean and makes the remaining comments more valuable to future readers.

main/src/main/java/se/hjulverkstan/main/jobs/S3OrphanFileCleanup.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Component
+public class S3OrphanFileCleanup {
+    private static final Logger logger = LoggerFactory.getLogger(S3OrphanFileCleanup.class);
+
+    private static final int BATCH_DELETE_THRESHOLD = 10;
+    private static final int BATCH_SIZE = 100;
```
jimmsi, line 20: Previously, I suggested using individual deletion for small numbers of orphaned files and batch deletion for larger sets. However, after reading Jona’s comment and reconsidering the tradeoffs, I now recommend always using batch deletion, even for small sets (e.g. 1–10 files). It simplifies the code by removing branching logic, it makes testing and maintenance easier, the S3 API supports batch deletion efficiently, even for small batches and we can still retain meaningful logs per batch if needed.

Cwejman (approved):
Looks good, no more comments

Cwejman:
I've approved this before, now it needs to be rebased and still one requested changed needs to be resolved. Please close and "merge" as soon as possible

PR #229, opened 2025-04-03
fix(api): Unable to create JAR file on windows

Daviduber69:
DotenvApplicationContextInitializer class would get errors on its projectRootPath when running mvn package (for getting JAR file) since the replacing of "/main", to empty string only works on Unix-based environments, but adding "\\main" makes it works on windows environments aswell.


Cwejman (changes requested):
This PR contains a lot more than the fix for building on windows. Clean it up check the commits as well there is a wip-commit included

Cwejman (approved):

PR #230, opened 2025-04-04
fix(web): Cannot search for vehicleCount or audit dates in Location 

### LINKED ISSUE #194: Cannot search for vehicle count, created/edited at in Location
take a look at:

web/src/root/Portal/PortalAdminLocations/AdminLocationsFilters.tsx

created and edited at is easy, they already exist as variables but vehicle count does not

Madasa-cell:
Resolves #194


Cwejman (changes requested):
No comment on code but the commit message could be more aligned with our guidelines.

That is: fix: What the problem is.

So perhaps

`fix(web): Cannot search for vehicle count or audit dates in Location`

Cwejman (approved):

Issue #231, opened 2025-04-04
Proxy backend in frontend dev server


Cwejman:
## Background

Discussions lately around our same-site option for cookies have settled in the following:

The current goal, to in AWS set up the backend on one subdomain and the front end on another mean that we need a more liberal security level on our auth cookies. 

And therefore opt to but the backend and frontend on the same site and proxy the backend so that it is reachable on `/api`.

With this change in architecture we need to reflect the same approach in the local development server that us using express.js and Vite's hot module reloading stuff.

## The tasks to accomplish is therefore to:

1. Update the .env.template and .env and set the VITE_WEB_API to `/api`.
2. Add a new variables to .env and .env.template, lets say VITE_WEB_API_PROXY and set it to the localhost url of the server.
3. Add a proxy to express.js to proxy requests at VITE_WEB_API to VITE_WEB_API_PROXY

## More thing

Maybe same site strict does not work for local development, if that is the case maybe add a flag if there isn't already one in spring to the application dev properties to disable same site.

## Comments

I propose to continue this work on the PR #209 as this issue makes the PR outdated but has some code in common.

Keep the commit that adds VITE_WEB_API but remove the commit that add CORS and create a new commit to remove CORS completely from application.properties and .env.template as it will no longer be needed

See #232 for reference on how to proxy with express.js

PR #236, opened 2025-04-04
chore(api): Api url moved to env-var

Daviduber69:


Daviduber69:
this is WIP for issue #231

jimmsi (approved):

jimmsi (approved):
Tried it out, works locally.

.env.template
```diff
@@ -1,7 +1,8 @@
-BUILD_WEB_SSR_API_USERNAME=christopher
-BUILD_WEB_SSR_API_PASSWORD=password
-VITE_WEB_DOMAIN=http://localhost:5173
-VITE_WEB_API=http://hjulverkstan:8080/v1
+VITE_BUILD_USERNAME=christopher
+VITE_BUILD_PASSWORD=password
+VITE_BACKEND_URL=http://localhost:8080/v1
+VITE_FRONTEND_URL=http://localhost:5173
```
Cwejman, line 4: Given that you rename the env vars here Daniel initial commit is override, better to push a series of commits to main that are aligned rather than overriding each other

main/src/main/java/se/hjulverkstan/main/service/CookieServiceImpl.java
```diff
@@ -50,28 +59,39 @@ public void refreshToken(HttpServletResponse response, String requestRefreshToke
...
                         "Refresh token is not in database!"));
+        RefreshToken refreshToken = refreshTokenService.findByToken(refreshedTokens.get("refreshToken"))
+                .orElseThrow(() -> new TokenRefreshException("Refresh token not found", "Invalid refresh token"));
 
-        setJwtCookie(response, refreshedTokens.get("accessToken"));
-        setRefreshCookie(response, refreshedTokens.get("refreshToken"));
+        User user = refreshToken.getUser();
+        UserDetails userDetails = new UserDetails();
+        List<String> roles = user.getRoles().stream()
+                .map(role -> role.getName().toString())
```
Cwejman, line 68: I think we need to review this code further, also the whole of this method, you've added complexity I'm not convinced about but dont have any further time at the moment to review. Lets get back to it in the afternoon or on Monday!

web/scripts/serve.js
```diff
@@ -0,0 +1,46 @@
...
+const rootPath = path.resolve(fileURLToPath(import.meta.url) + '../../..');
+config({ path: path.resolve(rootPath, '../.env') });
+
+const proxy = process.env.VITE_BACKEND_PROXY_SLUG;
+const api = process.env.VITE_BACKEND_URL;
+
+const app = express();
+
+const staticDir = `${rootPath}/dist/static`;
+fs.mkdirSync(staticDir, { recursive: true });
```
Cwejman, line 17: Why should the serve script create the dist folder if it does not exist? If there is not one there is not one, and the script should just fail with explicit reason

web/scripts/serve.js
```diff
@@ -0,0 +1,46 @@
...
+    target: api,
+    changeOrigin: true,
+    pathRewrite: (path) => path.replace(/^\/api/, ''),
+  }),
+);
+
+app.use(express.static(staticDir));
+
+app.use((req, res) => {
+  res.sendFile(path.join(staticDir, 'index.html'));
```
Cwejman, line 40: Perhaps it would be better to only return static dir if the url is not a file, so that if the requested url is /fonts/wrong-font-file.otf

It does not return the index.html

Cwejman (changes requested):

main/src/main/resources/application.properties
```diff
@@ -12,8 +12,6 @@ hjulverkstan.aws-s3.secret-key=${API_AWS_S3_SECRET_KEY}
 hjulverkstan.aws-s3.region=${API_AWS_S3_REGION}
 hjulverkstan.aws-s3.bucket-name=${API_AWS_S3_BUCKET_NAME}
 
-cors.allowedOrigins=https://hjulverkstan.org
```
Cwejman, line 15: After discussions in person we've decided to keep cors and keep it as an env var controlled value

main/src/main/java/se/hjulverkstan/main/service/CookieServiceImpl.java
```diff
@@ -30,9 +39,9 @@ public CookieServiceImpl(JwtUtils jwtUtils, RefreshTokenService refreshTokenServ
     public void createAuthenticationCookies(HttpServletResponse response, UserDetails userDetails) {
         String jwt = jwtUtils.generateToken(userDetails.getUsername());
         RefreshToken refreshToken = refreshTokenService.createRefreshToken(userDetails.getId());
-
-        setJwtCookie(response, jwt);
-        setRefreshCookie(response, refreshToken.getToken());
+        String sameSite = determineSameSite(userDetails);
+        setJwtCookie(response, jwt, sameSite);
+        setRefreshCookie(response, refreshToken.getToken(), sameSite);
```
Cwejman, line 44: From discussions in person:

Lets include support for refresh tokens for pipelines role by breaking out the somewhat biased approach with the creation of refreshedTokens map.


```java
RefreshToken currRefreshToken = refreshTokenService.findByToken(requestRefreshToken)
        .orElseThrow(() -> new TokenRefreshException(requestRefreshToken, "Refresh token is not in database!"));
refreshTokenService.verifyExpiration(refreshToken);

User user = currRefreshToken.getUser()
String jwt = jwtUtils.generateToken(user.getUsername());
RefreshToken nextRefreshToken = refreshTokenService.createRefreshToken(user.getId());

String sameSite = determineSameSite(user.getRoles()) // determine takes roles instead.
setJwtCookie(response, refreshedTokens.get("accessToken"), sameSite);
setRefreshCookie(response, refreshedTokens.get("refreshToken"),sameSite);
```

main/src/main/java/se/hjulverkstan/main/service/CookieServiceImpl.java
```diff
@@ -22,6 +25,12 @@ public class CookieServiceImpl implements CookieService {
     @Value("${saveChild.app.jwtRefreshExpirationMs}")
     private Long refreshTokenDurationMs;
 
+    public String determineSameSite(UserDetails userDetails) {
+        if(userDetails.getRoles().contains("ROLE_PIPELINE")){
```
Cwejman, line 29: Spaces is a bit of here, could be just a ternary as well.

web/scripts/serve.js
```diff
@@ -0,0 +1,48 @@
...
+    pathRewrite: (path) => path.replace(/^\/api/, ''),
+  }),
+);
+
+app.use(express.static(staticDir));
+
+app.use((req, res, next) => {
+  const reqPath = path.join(staticDir, req.url);
+  fs.stat(reqPath, (err, stats) => {
+    if (err || !stats.isFile()) {
```
Cwejman, line 37: If goes to a url and it is a file then express.static() should resolve it, so I think in this part of the code we already know there is no file. So just res.sendfile() is enough for this resolver

Cwejman (changes requested):
With the renaming of endpoints the dev scripts proxy and serve scripts proxy should reflect the new renamed endpoint. Also update the env.template to include v1 in the same commit as the changing of endpoints so that all commits in the PR work.

Cwejman (approved):

PR #240, opened 2025-04-09
Startpage MVP

daniel-svendsen:


web/public/trafficicon1.svg
Cwejman, line 1: Use something from lucide instead

web/src/client.tsx
Cwejman, line 1: Dont mount navbar and footer here

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
+import React, { CSSProperties, HTMLAttributes } from 'react';
+import { cva, VariantProps } from 'class-variance-authority';
+import { cn } from '@utils';
+import { LucideIcon } from 'lucide-react';
+
+const iconVariants = cva('flex-shrink-0', {
+  variants: {
+    variant: {
+      base: 'text-primary',
```
Cwejman, line 9: For coherence name default

Or name primary to be aligned with variable name

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
...
+Icon.displayName = 'CardIcon';
+
+//
+
+const titleVariants = cva(
+  'text-foreground font-inter z-10 font-semibold leading-tight',
+  {
+    variants: {
+      variant: {
+        base: 'text-h3',
```
Cwejman, line 43: text-h3 does not exist. Add text-foreground here instead.

And call the variant either default for coherence or aligned with the variable name foreground.

As mentioned before let's not use direct tailwind colors like text-white. The diesgn system is themeized so we can only use our established colors. Use text-background here instead and call the variant background

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
...
+  children: React.ReactNode;
+  className?: string;
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+}
+
+const Title: React.FC<CardTitleProps> = ({
+  variant,
+  children,
+  className,
+  as: Component = 'h2',
```
Cwejman, line 65: The title in cards should not be headings as they are not headings for sections of content but rather a piece of content. And anyway should not have varied sie. There is only one size of title in a cards. Style a piece of text instead.

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
...
+}
+
+const Image: React.FC<CardImageProps> = ({
+  src,
+  alt,
+  variant = 'background',
+  className,
+  ...props
+}) => {
+  if (variant === 'background') {
```
Cwejman, line 147: Use ternary instead to match with guidelines

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
...
+  );
+};
+Image.displayName = 'CardImage';
+
+//
+
+const cardBaseVariants = cva('flex overflow-hidden rounded-lg shadow-sm', {
+  variants: {
+    variant: {
+      base: 'h-auto flex-col gap-6 bg-white p-8',
```
Cwejman, line 185: Again with naming base, default here would be more aligned with the rest of the app

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
...
+  children: React.ReactNode;
+  className?: string;
+  src?: string;
+  alt?: string;
+  minHeight?: string | number;
+}
+
+const CardBaseRoot: React.FC<CardBaseProps> = ({
+  children,
+  variant = 'base',
```
Cwejman, line 215: Setting default argument here for variant is not needed

web/src/components/Card.tsx
```diff
@@ -0,0 +1,246 @@
...
+}) => {
+  const variantStyles: CSSProperties = {};
+
+  return (
+    <div
+      className={cn(cardBaseVariants({ variant }), className)}
+      style={variantStyles}
+      {...props}
+    >
+      {variant === 'imageAbove' ? (
```
Cwejman, line 229: No need for conditional rendering here, it just adds more styles, with is the purpose of CVA

web/src/components/Footer.tsx
Cwejman, line 1: Footer shop and footer text is not used by other components but are part of the components dir. Better to include them in the same file – the alternative should be to make a directory instead and contain them with the footer but I dont se the point here. It is not too much code to keep in one file, recommend to declare them bellow the Footer instead.

web/src/components/FooterShop.tsx
```diff
@@ -0,0 +1,66 @@
...
+      phone: '032423438',
+    },
+    {
+      name: 'Backa Shop',
+      address: 'Västra gården bakom 23, 443 14 Hisingsbacka',
+      phone: '032423438',
+    },
+    {
+      name: 'Biskopsgården Shop',
+      address: 'Femte spåret vid sidan 34, 423 11 Göteborg',
```
Cwejman, line 17: Lets use usePreloadedDataLocalised here before we close the PR

web/src/components/FooterText.tsx
```diff
@@ -0,0 +1,56 @@
...
+      className="flex flex-col items-center gap-4 text-sm text-gray-600
+        md:flex-row md:gap-6 md:text-base"
+    >
+      <div className="flex items-center gap-2">
+        <span>initiative by</span>
+        <Link
+          to="https://www.savethechildren.net"
+          target="_blank"
+          variant="link"
+          className="flex items-center gap-1 font-semibold text-gray-700
```
Cwejman, line 15: I think it makes more sense to make one link here, the one repeated three times as a component to then be reused in the footer. That will reduce the footprint by 300%

web/src/components/FooterShop.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <div
+      className="grid w-full grid-cols-1 gap-x-8 gap-y-10 text-center
+        sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-5 lg:text-left"
+    >
+      {shops.map((shop) => {
+        const addressParts = shop.address.split(',');
+        return (
+          <div
```
Cwejman, line 40: Same here as with footer text, better to encapsulate the repeated content as a component and reuse that in the footer instead. That will reduce footprint. The component is a slo a function which makes a natural closure for the `const addressParts...` code instead of keeping it in the map statement.

web/src/components/Navbar.tsx
```diff
@@ -0,0 +1,116 @@
...
+          sm:px-6 lg:px-8"
+      >
+        <div className="text-foreground text-xl font-bold">Hjulverkstan</div>
+
+        <nav className="hidden items-center gap-6 text-sm md:flex">
+          <Link to="/" className="hover:underline">
+            Home
+          </Link>
+          <Link to="/shops" className="hover:underline">
+            Shops
```
Cwejman, line 25: Is prettier running locally?

I would expect a link to be inline since it is not more than 100chars wide

web/src/components/Navbar.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <Link to="/stories" className="hover:underline">
+            Stories
+          </Link>
+          <Link to="/support" className="hover:underline">
+            Support
+          </Link>
+          <Link
+            to="/contact"
+            className="bg-primary text-primary-foreground hover:bg-primary/90
+              rounded-full px-4 py-2 text-sm font-medium no-underline
```
Cwejman, line 39: This style shot d be a button variant

web/src/components/Navbar.tsx
```diff
@@ -0,0 +1,116 @@
...
+            </Link>
+            <Link
+              to="/shops"
+              onClick={handleNavClick}
+              className="text-foreground block rounded-md px-3 py-2 text-base
+                font-medium hover:bg-gray-50"
+            >
+              Shops
+            </Link>
+            <Link
```
Cwejman, line 78: Good to see the mobile menu, but we never made any design for in figma.

It thin, when reviewing the code that perhaps the links to not need different styles in the mobile menu, just the contains layout is different, This way the code also is cleaner since the links can be reused thourgh conditional rendering, or a div surrounding the nav bar link in the navbar actually is the div that becomes the mobile menu.

web/src/components/PartnersSection.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <div className="mb-12 flex w-full items-center justify-between">
+        <h2 className="text-3xl font-bold">Our partners</h2>
+        <a
+          href="#"
+          className="flex items-center gap-2 text-sm text-gray-800 no-underline
+            hover:text-gray-900"
+        >
+          See all partners
+          <ArrowRight className="h-4 w-4" />
+        </a>
```
Cwejman, line 27: For now we have said to include all home page specific content in the home ecompoentn until we divide them out together, The parters section could therefore be there, also it ought to use the section component instead of repeating style for it

web/src/components/ShopAddress.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const ShopAddress: React.FC<ShopAddressProps> = ({
+  icon: Icon,
+  address,
+  isOpen,
+}) => {
+  return (
+    <div className="flex w-full justify-between pt-2 text-sm">
+      <div className="flex items-center gap-1">
+        <Icon size={16} className="flex-shrink-0 text-gray-600" />
```
Cwejman, line 22: Lets create the bullet component as discussed instead since this markup and style is to be reused in several place with different content

web/src/components/Statistics.tsx
```diff
@@ -0,0 +1,49 @@
...
+        'gap-x-8 gap-y-6 md:gap-x-12 lg:gap-x-16',
+        'py-10 md:py-16',
+        className,
+      )}
+    >
+      {stats.map((stat, index) => (
+        <div
+          key={index}
+          className="flex min-w-[100px] flex-col items-center text-center
+            md:min-w-[120px]"
```
Cwejman, line 32: Again we've said to move all home page specific content to the home component first before breaking it out gogether.

This is also a section and should use the section component

Cwejman (changes requested):

web/src/components/CardBase.tsx
Cwejman, line 1: Rename to CardDefault

web/src/components/CardBase.tsx
```diff
@@ -0,0 +1,52 @@
...
+  title,
+  body,
+  link,
+  linkLabel,
+  ariaLabel,
+  variant,
+  className,
+}) => {
+  return (
+    <Card.Base variant={variant} className={cn('flex flex-col', className)}>
```
Cwejman, line 29: flex col should probably be built into card base

web/src/components/CardBase.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Card.Icon icon={icon} />
+      <Card.Title>{title}</Card.Title>
+      <Card.Body>{body}</Card.Body>
+      <Button
+        asChild={!linkLabel}
+        variant="roundedPrimary"
+        className={cn('mt-auto', linkLabel ? '' : 'ml-auto h-8 w-8 p-0')}
+        aria-label={ariaLabel}
+      >
+        <Link to={link}>
```
Cwejman, line 39: Use link from button, it behaves the same as button

Cwejman: I also commented that maybe we should do a IconLink just like our IconButton so further simplify usage of links

web/src/components/ShopAddress.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <span
+          className="h-1.5 w-1.5 flex-shrink-0 rounded-full bg-red-500"
+          aria-hidden="true"
+        ></span>
+        <span>Closed</span>
+      </Badge>
+    ) : undefined;
+
+  return (
+    <Bullet icon={Pin} aside={badgeToShow} className={className}>
```
Cwejman, line 45: Lets skip shop adress as the is open content is to be reused outside of bullets later, and the verbosity of writing.

Skip aside as well as the badge can just be right next to the text

```typescript
<Bullet icon={...}>{address}<OpenBadge isOpen={...} className="pl-2" /></Bullet>
```

Cwejman (changes requested):
More comments after meeting at your desk and reviewing card templates

web/src/utils.tsx
```diff
@@ -218,3 +219,76 @@ export function matchDateWithoutTimestamp(
...
+
+  if (!hoursString) {
+    return false;
+  }
+
+  const timeParts = hoursString.split(/\s*-\s*/);
+  if (timeParts.length !== 2) {
+    console.warn(`Invalid hours format for ${currentDayKey}: ${hoursString}`);
+    return false;
+  }
```
Cwejman, line 255: Prompt to chatGPT

`Review this code and keep an inline and declarative style, focusing on simplicity and minimal code footprint`

Gave:

```typescript
const dayKeys: (keyof OpenHours)[] = ['sun', 'mon', 'tue', 'wed', 'thu', 'fri', 'sat'];

export function isShopOpen(openHoursData: OpenHours | null | undefined, date: Date): boolean {
  const hours = openHoursData?.[dayKeys[date.getDay()]];
  if (!hours) return false;

  const match = hours.match(/^(\d{2}):(\d{2})\s*-\s*(\d{2}):(\d{2})$/);
  if (!match) {
    console.warn(`Invalid format for ${dayKeys[date.getDay()]}: ${hours}`);
    return false;
  }

  const [_, oh, om, ch, cm] = match.map(Number);
  if ([oh, om, ch, cm].some(isNaN)) return false;

  const now = date.getHours() * 60 + date.getMinutes();
  const open = oh * 60 + om;
  const close = ch * 60 + cm;

  return close < open
    ? now >= open || now < close
    : now >= open && now < close;
}
```

web/src/root/Home.tsx
```diff
@@ -1,30 +1,389 @@
...
+  const { data } = usePreloadedDataLocalized();
+  const shops: Shop[] = data?.shop || [];
+
+  const statsData = [
+    { value: 628, label: 'Bikes repaired' },
+    { value: 500, label: 'Bikes saved' },
+    { value: 86, label: 'Bikes lent' },
+    { value: 20, label: 'Employees hired' },
+  ];
+
```
Cwejman, line 36: Even thought this data is not permanent and should be driven by webedit this data is static and can therefore be declareted outside of the component scope/closure

web/src/root/Home.tsx
```diff
@@ -1,30 +1,389 @@
...
+      <div className="-mx-4 h-px bg-[#D3D1EA]/50"></div>
+      <Section>
+        <SectionContent>
+          <div
+            className="flex flex-wrap items-baseline justify-center gap-x-24
+              gap-y-8 md:gap-x-56"
+          >
+            {statsData.map((stat, index) => (
+              <div
```
Cwejman, line 278: There should be two sections contents, one around the first content, the cards, and another around stories.

web/src/root/Home.tsx
```diff
@@ -1,30 +1,389 @@
...
+              >
+                See all stories
+                <ArrowRight className="h-6 w-6" />{' '}
+              </Link>
+            </div>
+
+            <div className=" flex flex-col gap-8 md:flex-row md:gap-8 lg:gap-8">
+              <div className="flex flex-col gap-8 md:basis-4/5 md:flex-row">
+                {stories
+                  .filter((s) => s.type === 'image')
```
Cwejman, line 194: there is no type on stories

web/src/root/Home.tsx
```diff
@@ -1,30 +1,389 @@
...
+                      body={story.bodyText}
+                      link={story.link}
+                      ariaLabel={story.ariaLabel}
+                    />
+                  ))}
+              </div>
+              <div className="flex flex-col gap-8 md:basis-1/4">
+                {stories
+                  .filter((s) => s.type === 'compact')
+                  .slice(0, 2)
```
Cwejman, line 211: Slice (2,4) here

web/src/root/Home.tsx
```diff
@@ -1,30 +1,389 @@
...
+      <Section>
+        <SectionContent>
+          <div className="flex flex-col">
+            <div
+              className="mb-[70px] flex flex-col items-start gap-4 sm:flex-row
+                sm:items-center sm:justify-between"
+            >
+              <div className="text-h2 text-foreground font-semibold">Shops</div>
+              <Button
+                variant="roundedMutedText"
```
Cwejman, line 236: Lets bake headings into section content. H2 will always be with a section content according to the existing design.

lets leave it optional as all section contents dent have heading, and lets use h2 instead of div.

And add optional props for link to and link label, if supplied will add the icon link, not a button with a link in it iwth and icon in it.

On mobile lets hade the link and show a link bellow the content instead.

Cwejman (changes requested):
From discussions in person

Cwejman (approved):

Issue #239, opened 2025-04-08
User role has unintended CRUD privileges


Daviduber69:
### ISSUE
Users can as of now in backend do CRUD operation, e.g. creating new employees in Postman
But cannot access or do any CRUDs in frontend, which is supposed to be the case.

Therefore we need to implement restrictions for User in the backend as well.

**a User should only be able to do CRUD in following endpoints**:

- /vehicles
- /image
- /ticket
- /customer

And GET requests in:
/employee
/location

ADMIN should have access to everything.
PIPELINE should only have access to LogIn and getall webedit: {all: webedit/get-all} endpoint. 

### IMPLEMENTATION
Spring has an annotation `@PreAuthorize`  and we could use this in the Controller layer.
PreAuthorize has a method `hasRole()` where you can assign a string value of say "ROLE_ADMIN".
(ROLE_ADMIN/USER/PIPELINEbecause that's how the enums are designed in ERole.java)

So for all endpoints explained above assign the correct Role to each of them in the Controller layer.

PR #242, opened 2025-04-11
fix(api): Roles not working in backend

Daviduber69:
#239 Roles now do what they are supposed to do. Reflecting frontend privileges. ROLE_PIPELINE which is found in a commit in #236 now has access to login and getAllWebEditEntitiesByLang only, and will have SameSite Strict from #236. Admin has all access, User has some access.


main/src/main/java/se/hjulverkstan/main/controller/webedit/ShopController.java
```diff
@@ -16,6 +17,7 @@
 
 @RestController
 @RequestMapping("v1/web-edit/shop")
+@PreAuthorize("hasRole('ROLE_ADMIN')")
```
Cwejman, line 20: Is it not best to use enums for this?

Cwejman (changes requested):

Cwejman (approved):

Issue #95, opened 2024-10-02
Public vehicle endpoint built int filtering


Cwejman:
The public endpoint for getting available bikes should improved.

It should filter on the following aspects built in (not optional)

- `isCustomerOwned` == `false`
- `status` == `available`
- every `ticketId` should not be a rental with `startDate` closer than `margin days available before rental start` constant.
If the startDate is starting within five days do not show it , otherwise show it
Where the `margin days available before rental start` constant is set to say `5` (days)

something query, shop{id} filtering showing bikes on that shop
Add new endpoint for public/shop/id, and use locationID to find out which is which and then show those vehicles bound to that shop.

PR #244, opened 2025-04-15
feat(api): Filtering on public vehicle endpoint

Daviduber69:
#95 The list of public vehicles now only includes vehicles that aren't customer owned, are available and aren't to be rented out within the coming 5 days. Also added new optional string for public/vehicle with ?locationId=n to see which of these filtered vehicles are available per location.

### DIFF
--- a/api/src/main/java/se/hjulverkstan/main/repository/PublicRepositoryImpl.java
+++ b/api/src/main/java/se/hjulverkstan/main/repository/PublicRepositoryImpl.java
@@ -1,0 +1,77 @@
+package se.hjulverkstan.main.repository;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.PersistenceContext;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+import se.hjulverkstan.main.model.TicketStatus;
+import se.hjulverkstan.main.model.TicketType;
+import se.hjulverkstan.main.model.Vehicle;
+import se.hjulverkstan.main.model.VehicleStatus;
+
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Optional;
+
+@Repository
+@Transactional
+
+public class PublicRepositoryImpl implements PublicRepository {
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    @Value("${rentalTimeMarginInDays}")
+    private int rentalTimeMarginInDays;
+
+    private List<Vehicle> findPublicAvailableVehiclesWithinMargin(Long vehicleId, Long locationId) {
+        LocalDateTime now = LocalDateTime.now();
+        LocalDateTime margin = now.plusDays(rentalTimeMarginInDays);
+
+        String baseQuery = """
+            SELECT v FROM Vehicle v
+            WHERE v.isCustomerOwned = false
+              AND v.vehicleStatus = :status
+              AND NOT EXISTS (
+                  SELECT t FROM Ticket t
+                  JOIN t.vehicles tv
+                  WHERE tv = v
+                    AND t.ticketType = :rentType
+                    AND t.ticketStatus != :closed
+                    AND t.startDate BETWEEN :now AND :margin
+              )
+        """;
+
+        if (vehicleId != null) {
+            baseQuery += " AND v.id = :vehicleId";
+        }
+        if (locationId != null) {
+            baseQuery += " AND v.location.id = :locationId";
+        }
+
+        var query = entityManager.createQuery(baseQuery, Vehicle.class)
+                .setParameter("status", VehicleStatus.AVAILABLE)
+                .setParameter("rentType", TicketType.RENT)
+                .setParameter("closed", TicketStatus.CLOSED)
+                .setParameter("now", now)
+                .setParameter("margin", margin);
+
+        if (vehicleId != null) {
+            query.setParameter("vehicleId", vehicleId);
+        }
+        if (locationId != null) {
+            query.setParameter("locationId", locationId);
+        }
+
+        return query.getResultList();
+    }
+     @Override
+     public List<Vehicle> findPublicAvailableVehicles(Long locationId) {
+         return findPublicAvailableVehiclesWithinMargin(null, locationId);
+     }
+
+    @Override
+    public Optional<Vehicle> findById(Long id) {
+        return findPublicAvailableVehiclesWithinMargin(id, null).stream().findFirst();
+    }
+}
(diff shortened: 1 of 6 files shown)


main/src/main/java/se/hjulverkstan/main/controller/PublicController.java
```diff
@@ -20,12 +20,16 @@ public PublicController(VehicleService vehicleService) {
     }
 
     @GetMapping("/vehicle")
```
Cwejman, line 22: Add shop id as an optional query param/string on the usual /vehicle endpoint instead

main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
```diff
@@ -72,18 +74,52 @@ public VehicleDto createVehicle(NewVehicleDto newVehicle) {
         return convertToDto(vehicle);
     }
 
+
     @Override
-    public GetAllVehicleDto getAllVehicles() {
+    public GetAllVehicleDto getAllPublicVehicles() {
+        List<Vehicle> allVehicles = vehicleRepository.findAll();
```
Cwejman, line 80: Better to add a method to the repository instead, nicer to do in sql.

main/src/main/java/se/hjulverkstan/main/controller/PublicController.java
Cwejman: Also rename to be locationId not shopId since this is a location not a web edit shop

main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
Cwejman: Lets hypothesise that we have the same repo method for both get vehicles and get vehicles by location id, since they will be under the same controller.

The service methods wraps the repo method and passes along the id, as null or string.

The repo method then either filters by it or not.

The repo should then match all the criteria thourgh the sql statement:

- is available
- not customer owned
- and with (constant) margin till next rental date

And if location is is passed filter on the location id as well.

Cwejman (changes requested):

main/src/main/java/se/hjulverkstan/main/repository/PublicRepositoryImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final int rentalTimeMarginInDays = 5;
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    private List<Vehicle> findPublicAvailableVehiclesNotRentedOutWithin5Days(Long vehicleId, Long locationId) {
+        LocalDateTime now = LocalDateTime.now();
+        LocalDateTime margin = now.plusDays(rentalTimeMarginInDays);
+
+        String baseQuery = """
+            SELECT v FROM Vehicle v
```
jimmsi, line 28: Using CriteriaBuilder would normally be a strong alternative to raw JPQL string queries. It provides several advantages like type safety, better IDE support (e.g. autocomplete and navigation), easier predicate composition, and more maintainable queries that are resilient to refactoring.

That said, based on the existing examples in the codebase (such as the native SQL used for S3 image URL operations), it seems we have intentionally opted for native SQL in certain cases when dealing directly with table names and cross-entity logic.

Still, even when sticking with JPQL strings, it would improve readability and reduce error risk to refactor the query construction to use StringBuilder instead of plain string concatenation. This helps structure conditional clauses more clearly and avoids subtle bugs related to spacing or missing keywords.

main/src/main/java/se/hjulverkstan/main/repository/PublicRepositoryImpl.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+
+@Repository
+@Transactional
+public class PublicRepositoryImpl implements PublicRepository {
+    private static final int rentalTimeMarginInDays = 5;
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    private List<Vehicle> findPublicAvailableVehiclesNotRentedOutWithin5Days(Long vehicleId, Long locationId) {
```
jimmsi, line 23: method name is a bit long and dependent on implementation details?  findAvailablePublicVehiclesFiltered could be an alternative with a comment explaining filtering criteria further, just a suggestion.

main/src/main/java/se/hjulverkstan/main/service/VehicleServiceImpl.java
```diff
@@ -72,18 +76,42 @@ public VehicleDto createVehicle(NewVehicleDto newVehicle) {
         return convertToDto(vehicle);
     }
 
+
     @Override
-    public GetAllVehicleDto getAllVehicles() {
+    public GetAllVehicleDto getAllPublicVehicles() {
```
jimmsi, line 81: It looks like the methods in `VehicleService` used by `PublicController` are only relevant to the public API and are not reused elsewhere in the system. Since they also rely solely on `PublicRepository`, this suggests they conceptually belong in a separate `PublicVehicleService`. Keeping them in `VehicleService` might imply shared business logic across contexts, but that's not the case here. Extracting them would improve clarity, follow SRP more closely, and make the system easier to maintain and understand.

I guess there could be other opinions, and even if I'm right, it might not be crucial ... thoughts on this?

main/src/main/java/se/hjulverkstan/main/controller/PublicController.java
```diff
@@ -20,12 +17,16 @@ public PublicController(VehicleService vehicleService) {
...
-    public ResponseEntity<GetAllVehicleDto> getAllVehicles() {
-        return new ResponseEntity<>(vehicleService.getAllVehicles(), HttpStatus.OK);
+    public ResponseEntity<GetAllVehicleDto> getAllPublicVehicles(@RequestParam(required = false) Long locationId) {
```
jimmsi, line 20: Is Public Controller linked to vehicles only, in that case it maybe should be renamed to better reflect its purpose? Although maybe out of scope for this issue. Same comment for PublicRepository of course

jimmsi (commented):
Some comments for discussion

api/src/main/java/se/hjulverkstan/main/repository/PublicRepositoryImpl.java
```diff
@@ -0,0 +1,73 @@
...
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Optional;
+
+@Repository
+@Transactional
+public class PublicRepositoryImpl implements PublicRepository {
+    private static final int rentalTimeMarginInDays = 5;
+    @PersistenceContext
+    private EntityManager entityManager;
```
Cwejman, line 21: Config value rentalTimeMarginInDays is fine but make a config value to be controlled from application.properties.

But rename findPublicAvailableVehiclesNotRentedOutWithin5Days to `findPublicAvailableVehiclesWithinMargin`

Cwejman (approved):

PR #246, opened 2025-04-16
feat: Add phone number field to Shop entity, DTOs, and types.

a93a:
- Added phoneNumber field to Shop model entity.
- Updated all relevant Shop DTOs to include phoneNumber field.
- Add necessary changes to service layer to handle the new field during mapping and persistence.
- Added phoneNumber field to shop type in frontend.
- Updated the database seed table to have the value for field phone_number.

### DIFF
--- a/api/src/main/java/se/hjulverkstan/main/dto/webedit/NewShopDto.java
+++ b/api/src/main/java/se/hjulverkstan/main/dto/webedit/NewShopDto.java
@@ -38,4 +38,7 @@ public class NewShopDto {
 
     @NotBlank(message = "Body text in at least one language is required for creating a shop")
     private String bodyText;
+
+    @NotNull(message = "A phone number is required")
+    private String phoneNumber;
 }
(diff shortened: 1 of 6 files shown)


Daviduber69:
As of now phonenumbers are currently just required strings in frontend, i have no idea if we are to implement any validations there like we have with personal id numbers, but those seem like they should be stricter and validated more than a phone number. and having a seperate phonenumber class seems cool and all but it might be a bit overkill. But having validation in backend and frontend that a phonenumber can only be valid (i.e x digits long, and all that ) might be a good idea.

Daviduber69 (commented):
As of now phonenumbers are currently just required strings in frontend, i have no idea if we are to implement any validations there like we have with personal id numbers, but those seem like they should be stricter and validated more than a phone number. and having a seperate phonenumber class seems cool and all but it might be a bit overkill. But having validation in backend and frontend that a phonenumber can only be valid (i.e x digits long, and all that ) might be a good idea.

But since phonenumber is used in shop, customer and employee as required strings. Having a phonenumber class might not be such a bad idea after all.
Or just add regex pattern validation to all 3 classes, might not be bad either.

But your task was to add phonenumber field to Shop, which you have done, so maybe that is a different task.

Cwejman:
Closing this as it is now too stale unfortunately, a new PR has to be made as this is from a fork.

PR #250, opened 2025-04-22
Refactored code for Infra

ssalmansattar:
Pull request with all the files for infra front end and backend


Cwejman:
I decided to take a session with the code and implement some lingering ideas, mainly around decoupling environment variables from the stack.

I'll review briefly what I've done:

##1

First of prettier was run, which is positive but that means that almost every line of code is updated, increasing the footprint of my commit. My appologies for this but I'll add comments about all my changes:

##2

I've created a config.ts file which has a factory for creating a config object, this config object is then passed to all stacks and sub stacks and thus simplifies and centralizes the environment related configuration.

My intention here was to import environment variables for each environment based on prefixes like `DEV_` etc but I needed up not needing any environment specific env vars at all. This was solved by:

##3

Create a global stack and a local stack separately. This was so everything can be referenced when it comes to route53 things.

Check it out, it should be self validating with almost nothing needed to be hardcoded.

##4

With two stacks the file structure changed a little, in order to prevent a too nested structure I took the liberty and flattened the stack directories contents and used the typescript standard of index.ts.

##5

I added a domain to be generated for ssh access and therefore remove the exposing of the ec2 instance ip.

---

Well that is a little bit of helpful info on my work. But I think you will have no issues just reading thorugh the code directly to discern the changes. All in all the point was to remove all hardcoding and come up with a simple approach to propagation of configuration.

Unfortunately I have not had time to test it and deploy it so truthfully it is dead code. If it has more then three lines that need adjusting I'll buy fika!

Would you be able to review it tomorrow, I'm not sure I will find time to deploy myself, if I get time I'll get in touch.

Looking forward to your input.

---

One question about your code, the assets uploading to the site bucket looks like unfinished work?

Should it upload to the site bucket and does that code work, does it actually deploy docker compose file to the instance?

Curious to discuss this further when we have time.

Big thank you for all the help, I'm happy to get to work with it myself as well finally, even though it is nighttime....

Cheers,
Jona

ssalmansattar:
Assets are uploaded to site bucket in file frontend-stateless.ts line 130 as below
new s3deploy.BucketDeployment(
      this,
      `${props.config.resourcePrefix}DeployWebsite`,
      {
        sources: [s3deploy.Source.asset('./assets')],
        destinationBucket: props.siteBucket,
        role: deploymentRole,
      },
      
      
      We can test it by uploading new chunk of files, delete bucket is not enabled so everytime we need to manually delete bucket.

ssalmansattar:
Docker compose, docker, enviornment variables and binary for backend are uploaded manually.
Afterwards we need to compose up file, but docker is installed through cdk in ec2 instance.

Cwejman (approved):
Quite a lot of changes has been made since the comments in this thread last, please review the commits for insight into what the PR does.

PR #251, opened 2025-04-23
Feature/dynamic routes from webedit data

jimmsi:
This task involves extending the custom Static Site Generation (SSG) system to support dynamic routes, using web edit data to define which paths should be rendered both at build time and runtime.

The work consists of the following commits:

1. Support dynamic routes by adding slug to Shop
2. Add support for dynamic routes
3. Isolate AuthProvider to only wrap CSR portal routes

Regarding (2) it's worth noting that I also verified that nested dynamic segments like /shops/:city/:shop works, by hardcoding test values into the dynamicSegments array and confirming correct HTML generation during the build.


Daviduber69:
we've already talked about it in person but im writing it down for remeberance:
the build script does not work with the current implementation of createRoutes.

web/scripts/build.js
```diff
@@ -64,43 +66,58 @@ const buildRoute = ({ path, title, isSSR, data }) => {
...
         isSSR: true,
         data,
-      }),
-    );
+      });
+    });
   });
 
+  // Render default fallback (e.g. /sv/...)
   console.log('[INFO]: Rendering routes for root');
```
daniel-svendsen, line 112: log

web/scripts/build.js
```diff
@@ -64,43 +66,58 @@ const buildRoute = ({ path, title, isSSR, data }) => {
...
  * Build SSR routes in each locale and build at the root with default locale.
  * For more info see [Static Site Generation Strategy](link).
  */
 
 try {
   const data = await getDataForPreloadingServerSide(process.env);
+  const { ssr: routesSSR, csr: routesCSR } = createRoutes(data);
+
+  // Render CSR routes (e.g. Portal)
+  console.log('[INFO]: Rendering CSR routes');
```
daniel-svendsen, line 95: log

daniel-svendsen:
maybe add a Unique field for slug column and ShopServiceImpl should also handle potential conflicts if a user tries to create/update a shop with a slug that already exists.

Also might add useExpandRoute to utils? the function is created in 2(?) places and maybe more when the code gets bigger.

Issue #25, opened 2024-06-07
WebEdit Stories


Cwejman:
This taks does not include adding stories (previously named blog) to the webedit frontend.

BACKEND
- Create Stories classes in each respective webedit folder (i.e. webedit/Blogs - controller, dto, model, repository, service)
- Include the stories in the get all web edit entities by lang endpoint

DATA/API
- Update the typing for the get all web edit entities by lang

Cwejman:
Marked as unrefined until bullet point instructions from events is copied here.

PR #252, opened 2025-04-23
feat(api): Backend support for WebEdit Stories

Daviduber69:
#25 Implementing columns, DTOs, repository, service and controllers for WebEdits Stories. Full CRUD. Also adding Story in API on web edit entities by lang.

### DIFF
--- a/api/src/main/java/se/hjulverkstan/main/controller/webedit/StoryController.java
+++ b/api/src/main/java/se/hjulverkstan/main/controller/webedit/StoryController.java
@@ -1,0 +1,50 @@
+package se.hjulverkstan.main.controller.webedit;
+
+import jakarta.validation.Valid;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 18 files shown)


main/src/main/java/se/hjulverkstan/main/model/webedit/Story.java
```diff
@@ -0,0 +1,32 @@
...
+@ToString
+@EqualsAndHashCode(callSuper = true)
+public class Story extends Auditable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "text_type")
+    private TextType textType;
```
jimmsi, line 25: This field (textType) is only used in method mapToDto in service layer, i guess it should be used elsewhere as well...?

Cwejman (changes requested):
Otherwise no comments, the service layer could be refined but is just a duplicate of shops now, such discussions are of a bigger scope

Cwejman (approved):

PR #253, opened 2025-04-23
feat: Signout not to use ID in parameter

a93a:
Following changes were made:

- Change in frontend, remove id as parameter in calls to logout from `components/Auth.tsx` and `components/Auth.tsx`
- In backend:
- Change `JwtUtils `to have ID as claims to later be extracted, and extraction method
- Change in `AuthController `to extract ID from JwtUtils 
- Change in `CookieServiceImpl` to handle ID in method calls
- Move duplicate in `AuthController` to private method.
- Add `SecurityContextHolder.clearContext()` after signing the user out, in `logoutUser()`, to clear security context. While this is optional and not necessary for stateless APIs, it’s not harmful as long as you're not relying on sessions and can clear up leftover data. Could be useful when doing logging to get cleaner logs, so that old data does not get leaked in logging? Or was this intentionally left out per Hjulverkstan design?

Thoughts and suggestions:

- Having a background in Kotlin with stricter _null safety_, I suggest we could use  `Optional<T>` to handle _nullable_ objects safer. 

  For example `getToken()`...
  
  ```java
  private String getToken(HttpServletRequest request, String type) {
          String token = null;
  
          if (request.getCookies() != null) {
              token = Arrays.stream(request.getCookies())
                      .filter(cookie -> type.equals(cookie.getName()))
                      .findFirst()
                      .map(Cookie::getValue)
                      .orElseThrow(() -> new MissingArgumentException(type + " token"));
          }
  
          return token;
      }
  ```
  ...could be changed to...
  
  ```java
  private Optional<String> getToken(HttpServletRequest request, String type) {
      if (request.getCookies() == null) {
          return Optional.empty();
      }
  
      return Arrays.stream(request.getCookies())
              .filter(cookie -> type.equals(cookie.getName()))
              .findFirst()
              .map(Cookie::getValue);
  }
  ```
  
  ...and in `AuthController.java`...
  
  ```java
  String refreshToken = null;
          try {
              refreshToken = getToken(request, "refreshToken");
          } catch (Exception e) {
              return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(new MessageResponse("Cookies are empty or null"));
          }
  ```
  
  ...could...
  
  ```java
  Optional<String> refreshTokenOpt = getToken(request, "refreshToken");
  
      if (refreshTokenOpt.isEmpty()) {
          return ResponseEntity.status(HttpStatus.BAD_REQUEST)
                  .body(new MessageResponse("Cookies are empty or refresh token not found"));
      }
  ```
  
  Could changes of this kind be of use for better null safety, in the future?

- There is some _business logic_ in `AuthController`. Should this  be delegated to Service layer instead? Example, extracting tokens and data from tokens.
- Should we validate the token is not expired in `logoutUser()` before trying to logout the user, or is this 100% unnecessary?

### DIFF
--- a/main/src/main/java/se/hjulverkstan/main/controller/AuthController.java
+++ b/main/src/main/java/se/hjulverkstan/main/controller/AuthController.java
@@ -1,19 +1,23 @@
 package se.hjulverkstan.main.controller;
 
+import io.jsonwebtoken.JwtException;
 import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import jakarta.validation.Valid;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.security.authentication.InsufficientAuthenticationException;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.web.bind.annotation.*;
 import se.hjulverkstan.Exceptions.MissingArgumentException;
 import se.hjulverkstan.Exceptions.TokenRefreshException;
 import se.hjulverkstan.main.dto.MessageResponse;
 import se.hjulverkstan.main.dto.auth.LoginRequest;
 import se.hjulverkstan.main.dto.auth.UserDetails;
+import se.hjulverkstan.main.security.jwt.JwtUtils;
 import se.hjulverkstan.main.security.services.AuthService;
 import se.hjulverkstan.main.service.CookieService;
 import se.hjulverkstan.main.service.CookieServiceImpl;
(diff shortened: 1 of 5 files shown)


main/src/main/java/se/hjulverkstan/main/controller/AuthController.java
```diff
@@ -65,11 +66,29 @@ public ResponseEntity<?> refreshToken(HttpServletRequest request, HttpServletRes
...
 
-    @PostMapping("/signout/{id}")
+    @PostMapping("/signout")
     @PreAuthorize("hasRole('ROLE_ADMIN') or hasRole('ROLE_USER')")
-    public ResponseEntity<?> logoutUser(@PathVariable Long id, HttpServletResponse response) {
-        MessageResponse messageResponse = authService.signOut(id);
+    public ResponseEntity<?> logoutUser(HttpServletRequest request, HttpServletResponse response) {
+        String accessToken;
+        try {
+            accessToken = getToken(request, "accessToken");
```
Cwejman, line 74: While the usage of JwtUtils mean the controller is relieved of certain depth of implementation I still think that the controller takes too much responsibility here, of where the current logic would be disposed there still another important factor to take into account:

Here the accessToken is the source truth, and the accessToken is already parsed by the AuthTokenFilter and if successful parses the token and sets the spring security context with the principal UserDetailsImplementation which is populated by the UserDetailsServiceImplementation connected in the WebSecurityConfig and used by the AuthTokenFilter.

Now I'm not sure if our approach to spring security is correct, some parts I'm sceptical about given how spring security works.

The spring security layer is founded upon how it works and there's is quite good videos and documentation about it.

The principal is important here, it is already populated by the filters before the auth controller is reached and thus all information is already available if the token was deemed valid. So instead of implementing logic here in the controller the signOut method in authService does need a userId, it could in it self just access the principal directly, and get the id from there.

Looking deeper at the code it the spring security context can just be reset regardless of id, the id is just for removing the refresh token.

So the authService upon sign out could just 

```java
        Integer userId = (UserDetailsImplementation) SecurityContextHolder.getContext().getAuthentication().getPrincipal().getId();
```

Before invoking invoking the `deleteByUserId()`

Cwejman (changes requested):

Cwejman:
Hi Ali,

Same here, some work left. Unless you insist to resolve it yourself, someone in the team can finish it.

PR #254, opened 2025-04-24
fix(api): Authentization and web edit get all entities by lang error

Daviduber69:
More sensible leisure on roles


Daviduber69:
outdated

PR #255, opened 2025-04-24
feat(): Convert vehicles to customer owned on donate

### LINKED ISSUE #82: Donate ticket should convert vehicles to customer owned
Creating a donate ticket should prompt the user with a confirmation dialog.

Are you sure you want to give away these bikes? This action can not be undone. All vehicles will be converted to customer owned and archived.

### Frontend

1.

A dialog should prompt the user  when clicking *create* on a `donate` ticket:

> **Donate vehicles**
> *This will archive all vehicles and make customer owned: *
> "list of reg tags"
> `cancel` `yes`

In order to technically implement this we could, without having to modify the `<PortalForm />` component inject into the promise chain that is the mutation function and use the openDialog from #83 in it:

```typescript
  const vehiclesQ = useVehiclesQ();

  ...

  const { openDialog } = useDialogManager();
  
  const createTicketMaybeWithDialog = (body: CreateTicketParams) =>
    body.type !== TicketType.DONATE
      ? createTicketM.mutateAsync(body);
      : new Promise((resolve, reject) => openDialog(
        <ConfirmDialog
           title="Donate vehicles" 
           onConfirm={() => resolve(createTicketM.mutateAsync(body)}
           onClose={() => reject('closed')}>
         >
           This will archive all....
           <BadgeGroup badges={map over vehicleIds and find the regTags from vehiclesQ and map into a regTag badge}
         </ConfirmDialog>
        )
     // If rejected by closing the dialog the PortalForm should not recieve this as an error.
     }).catch((err) => err !== 'closed' && Promise.reject(err))

       
  // then use this function instead of createTicketM.mutateAsync that is passed to PortalForm
```

In this example a confirm dialog component is used, this will have to be created.

### Backend

Update the create ticket service, to if the ticket is a donate, convert all vehicles to customer owned and mark status as archived.

Madasa-cell:
When creating a donate ticket, the selected vehicles are now converted to customer owned, their regTags are nulled and then they are archived.

Resolves #82

### DIFF
--- a/web/src/root/Portal/PortalShopTickets/index.tsx
+++ b/web/src/root/Portal/PortalShopTickets/index.tsx
@@ -27,6 +27,9 @@ import {
   VehicleShortcutAction,
   VehicleShortcutLocationState,
 } from '../PortalShopInventory/ShopInventoryActions';
+import { useDialogManager } from '@components/DialogManager';
+import { CreateTicketParams } from '@data/ticket/api';
+import ConfirmConvertDialog from '@components/ConfirmConvertDialog';
 
 //
 
@@ -66,6 +69,23 @@ export default function PortalShopTickets({ mode }: PageContentProps) {
 
   const columns = useColumns();
 
+  const { openDialog } = useDialogManager();
+
+  const createTicketMaybeWithDialog = (body: CreateTicketParams) =>
+    body.ticketType !== TicketType.DONATE
+      ? createTicketM.mutateAsync(body)
+      : new Promise((resolve, reject) => {
+          openDialog(
+            <ConfirmConvertDialog
+              title="Converting vehicles with donate ticket"
+              description="You are donating vehicles owned by Hjulverkstan. This will archive all selections and cannot be undone."
+              onConfirm={() => resolve(createTicketM.mutateAsync(body))}
+              onClose={() => reject('closed')}
+              vehicleIds={body.vehicleIds}
+            />,
+          );
+        });
+
   return (
     <DataTable.Provider
       key="tickets"
@@ -104,7 +124,7 @@ export default function PortalShopTickets({ mode }: PageContentProps) {
               error={ticketQ.error || locationsQ.error}
               isSubmitting={createTicketM.isPending || editTicketM.isPending}
               saveMutation={editTicketM.mutateAsync}
-              createMutation={createTicketM.mutateAsync}
+              createMutation={createTicketMaybeWithDialog}
             >
               <ShopTicketsFields />
             </PortalForm>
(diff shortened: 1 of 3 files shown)


Daviduber69 (commented):
I tried it and it works as described.

Im just curious about one thing:

image


Why include` `#${v.id}`` in this badgegroup at all, since you can only donate hjulverkstan owned vehicles anyway? (regTag)
It's already forbidden in both back and frontend.

web/src/root/Portal/PortalShopTickets/index.tsx
```diff
@@ -66,6 +72,50 @@ export default function PortalShopTickets({ mode }: PageContentProps) {
...
+                      v,
+                    ): v is {
+                      id: string;
+                      regTag: string;
+                      vehicleType: VehicleType;
+                    } => !!v,
+                  )
+                  .map((v) => ({
+                    label: v.regTag,
+                    variant: 'secondary',
```
Cwejman, line 111: Break out the creation of the badges to a separate function could be a good way to reduce the footprint and depth of this createTicketMaybeWithDialog function.

Since the creation of the badges is completely pure and only takes on parameter it is a good candidate for breaking out into a pure function. Would make the createTicket... easier to read.

Cwejman (changes requested):

Daviduber69 (commented):
All seems to work!
I'm just wondering if the purpose of onClose (pressing Cancel) was supposed to have an Error Toast pop up? `Failed to ${verbLabel} the ${dataLabel}.`

It doens't seem like there is any inherit code to make the errorToast appear in the code, but maybe some other component uses this and makes it appear onClose(?)

But seeing as the Task suggested you  `onClose={() => reject('closed')}>` , which you have, I don't see how you have done anything wrong.

But it feels kinda weird to get an Error saying Something went Wrong, when you just pressed Cancel manually.

Madasa-cell:
> All seems to work! I'm just wondering if the purpose of onClose (pressing Cancel) was supposed to have an Error Toast pop up? `Failed to ${verbLabel} the ${dataLabel}.`
> 
> It doens't seem like there is any inherit code to make the errorToast appear in the code, but maybe some other component uses this and makes it appear onClose(?)
> 
> But seeing as the Task suggested you `onClose={() => reject('closed')}>` , which you have, I don't see how you have done anything wrong.
> 
> But it feels kinda weird to get an Error saying Failed to Do Thing, when you just pressed Cancel manually.

This happen because the rejection from `onClose={() => reject('closed')}` triggers the .catch on createMutation in onCreate() from the PortalForm.

```typescript
createMutation(transformBodyOnSubmit(body))
        .then((res: any) => {
          navigate('../' + res.id);
          toast(
            createSuccessToast({
              verbLabel: 'create',
              dataLabel: `${dataLabel}`,
              id,
            }),
          );
        })
        .catch((err: any) => {
          console.error(err);
          toast(
            createErrorToast({
              verbLabel: 'create',
              dataLabel: `${dataLabel}`,
            }),
          );
        });
```

We are currently treating cancel like an actual error. This is why we are getting a toast error even if nothing bad actually happened, we are also getting an error in the console.

Cwejman (approved):

Issue #34, opened 2024-06-07
Change password in Admin/Users


Cwejman:
One should be able to set a new password for users in the Admin. Currently only possible to set during creation.

## Frontend

There should be no column for password (as it currently is as of writing), if the admin user wishes to change a password they have to press the row of the user whose password they wish to change. A side menu will pop up for editing (already implemented), and in it there should be a password field that is censored (********) but changeable, this can be achieved with a placeholder field on input. The password should only be censored during edit mode but not during create mode, during read mode the field shouldn't be visible. Furthermore, there should be additional password field to confirm that the password written in the first field is similar (==) to the password in the second field. This second password field should only be visible during create and edit mode, it should always be visible in create mode and only turn visible in edit mode if the password field has been changed.

Changes will at least be done in /web/src/data/user/form.ts since that is where the create and edit form of the Admin/Users page is located.

Bonus: Add a standard to the password, IE it has to have a length of minimum 8 letters, 2 numbers, 2 special characters.

--------

Note: This Feature may have to be refined again in the future once Bug #48 has been fixed and the user information is once again visible. The current refinement is based on assumptions made from how the other pages look and from how user/form.ts looks. This should be correct, so this note is just here to be sure.

Semi-Blocked by #48, if you create a new User you can edit that user but nothing more.

Is blocking #146

## Backend

The Backend seems to have been done, password gets updated through updateUser in /main/src/main/java/se.hjulverkstan/main/service/UserServiceImpl.java. That was a lot of mains. There shouldn't be any work to be done here.

Cwejman:
Hmm, I see two approaches to this, with different technical paths and different UX.

When a password is updated it should be repeated, right? To prevent misspellings.

So it could be done in the edit/create panel and not visible in read. but the validation logic would have to be made, on create both fields are required and compared to each other in edit if one fields is edited then both need to be.

(bonus: we can then add a standard to the password, length, capital letters etc...)

The other approach would be from the actions menu. a reset password action that opens up a modal. but, hold up...

When creating a user the password would have to be in the edit panel anyway... Strike the modal idea then.

But a specification on the validation, like I mention above when should it be required and not is good to settle before implementation.

Cwejman:
[...]

PR #259, opened 2025-05-06
feat(web): Change password in Admin/Users

Daviduber69:
#34 You can now change password for users and also implementing a Confirm-Password-Field so the password provided is correct and remembered by the user. You cannot paste in the repeat password field for more security. When in edit mode before you type anything in the Change password field, you do not see the Repeat password field, but as soon as you start typing in Change password, then the repeat password field will pop up.


daniel-svendsen (approved):
works on portal

Daviduber69:
I did not limit the password creation with annoying safety measures such as 10 chars minimum, 2 special chars, and 2 capital chars because of experience at companies with logins for employees where you just want a quick easy password to log in with.

web/src/data/user/form.ts
```diff
@@ -4,23 +4,32 @@ import { AuthRole } from '@data/auth/types';
...
+      password:
+        mode === Mode.CREATE ? reqString('Password') : z.string().optional(),
+      passwordrepeat:
+        mode === Mode.CREATE
+          ? reqString('Matching password')
+          : z.string().optional(),
+    })
+    .refine((data) => data.password === data.passwordrepeat, {
+      path: ['passwordrepeat'],
+      message: 'Passwords must match!',
```
jimmsi, line 34: "Passwords must match!" appears before the user has actually tried to confirm their password. This message should probably only appear after the user has made an attempt to re-enter their password and failed.
Perhaps the best approach is to display the message when the user clicks Save?

Daviduber69: I didnt quite make it like this , but I redid it so that the Save button is greyed out and the message is on the Save button instead of under the form whenever ´the  PWs didnt match.

web/src/components/DataForm/DataFormProvider.tsx
```diff
@@ -90,8 +90,13 @@ export const Provider = U.withLobotomizer(
...
     const { toast } = useToast();
 
+    const omitPassword = (data: Partial<D>): Partial<D> => {
+      const { password, passwordrepeat, ...rest } = data as any;
+      return rest;
+    };
+
     const [body, setBody] = useState<Partial<D>>(
-      mode === Mode.CREATE ? initCreateBody : (data ?? {}),
+      mode === Mode.CREATE ? initCreateBody : omitPassword(data ?? {}),
```
Cwejman, line 99: Why omit password from data, password should never be filled on a get request right? When transitioning from create to read I think the form is lobotomised as well so no need to omit between the modes if the password is not available form the get request.

web/src/data/user/form.ts
```diff
@@ -4,23 +4,38 @@ import { AuthRole } from '@data/auth/types';
...
+export const userZ = (mode: Mode) =>
+  z
+    .object({
+      roles: z
+        .array(z.nativeEnum(AuthRole, isReq('User role is required')))
+        .min(1, 'At least one role is required'),
+      username: reqString('Username'),
+      email: reqString('Email').email({
+        message: 'The email is not a valid email address',
+      }),
```
Cwejman, line 24: Looks like a reasonable approach conveying it to a function, cant see any better approach at the moment, for coherency and for good naming could be prefixed with `to` or `create` which I almost always to, in this case it make it more clear as well that this schema is a function since all the others are not.

Cwejman (changes requested):

web/src/components/DataForm/DataFormProvider.tsx
Daviduber69: the form was filled with the whole encrypted password when editing, so that's why, but i can find an alternate way to do it without omitting the password mayhaps perchance :3

PR #262, opened 2025-05-09
Services draft wipperino swaggerino

Daviduber69:


Daviduber69:
Check out the 3rd commit, i think that one is the closest to correct.

PR #264, opened 2025-05-13
Investiagea broke web pipeline job

Cwejman:


Cwejman:
It was not the most brilliant approach that we have to create PRs to run the pipeline of non main branches, but this PR fulfilled its purpose. Closing it now.

Issue #274, opened 2025-05-22
Brand optional / unknown


Daviduber69:
Make `Brand `optional on creation/edit of vehicles

### FRONTEND

in `data/vehicle/form.ts`  you can see that the field brand isReq. 

Change that.

Alternatively you can create a new Brand type in the enum called `UNKNOWN` but seeing as there is already a brand called  `OTHER` I don't think this is as good of an idea as making Brand optional.

### BACKEND

You also have to change this in the Backend since in the DTO layer Brand is set as mandatory.

OmarRdz12:
Hello, I would like to help with this issue.

Cwejman:
Hi, @OmarRdz12. How nice to see a new contributor. You are of course welcome to proceed.

Go ahead and make a fork, we have a check list in the main readme to read up on please make sure you go through this so that you are up to speed on our guidelines, how to setup a local environment etc, before you make a PR.

If you find the documentation lacking or you have input / question along the way, feel free to write here.

/ Jona

Cwejman:
Closed 5months ago by #288

PR #288, opened 2025-05-26
fix: make brand optional

OmarRdz12:
# Description
The brand was made optional, both in the form.ts and in the respective dtos.

## Explanation
- The notations in DTOs that made the brand field mandatory when creating objects have been removed.
- The zod object that defined the Vehicle has been modified to accept brand as optional.
- The Select field has been modified to allow deselecting an option and, if this option is enabled, to display a different placeholder.

Solves #274

### DIFF
--- a/api/src/main/java/se/hjulverkstan/main/dto/vehicles/EditVehicleBikeDto.java
+++ b/api/src/main/java/se/hjulverkstan/main/dto/vehicles/EditVehicleBikeDto.java
@@ -24,7 +24,6 @@ public class EditVehicleBikeDto extends EditVehicleDto {
     @NotNull(message = "Brake type is required")
     @JsonFormat(shape = JsonFormat.Shape.STRING)
     private VehicleBrakeType brakeType;
-    @NotNull(message = "Brand is required")
     @JsonFormat(shape = JsonFormat.Shape.STRING)
     private VehicleBrand brand;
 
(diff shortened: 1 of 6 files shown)


Cwejman (changes requested):
Code looks good, I will give it a test in the office tomorrow and get back to you.

Two minor things:

- The .idea directory has from some time back started to be included in commitable files unfortunately, which it shouldn't. So please remove the file from the commit.

- Also I don't think your work is bug fix but an added feature to the application, you're welcome to name your commit `feat: Make brand optional` instead.

Thank you for your contribution,
Jona

OmarRdz12:
Hello, thanks for your feedback, I did the changes, anything else you need, you can tell me.

OmarRdz12:
Hi, I have a question. What do you mean by the backend isn't updated? According to the tests I've run, both the frontend and backend respond well to the optional flag. If there's a specific part, let me know, and I'll be happy to make the necessary changes.

I made the change to the select option; now, when I click on the previously selected option, it will be removed and can be left empty again. I also changed the placeholder so that if the select option behaves this way, it now displays the message "No ___ selected."

I discovered the project while searching for contributions to open source projects. Yes, I managed to run the project locally, and I think the only feedback I can give on the process is that this part
"Click on 'Go to Portal' to log into the Portal."
That option never appeared; I had to navigate to the path via the browser URL.

I'm waiting for your review. Best regards.

OmarRdz12:
Hi @Cwejman I'd like to know if there's anything I should do or if I should just wait for the merge?

Cwejman (approved):

Cwejman:
Ah, yes of course. You are the first external contributor, so that is on me. Closing.

Thank you,

If there is something in particular you'd like to explore or learn, feel free to reach out and we can see if there are any matching tasks. There is more to the project than the backlog that is available as of now or if you have ideas of your own.

Cheers,
Jona

PR #302, opened 2025-06-23
feat: Notify customer by SMS when case status sets to compelte

### LINKED ISSUE #276: Automail / sms on repair completion
Hjulverkstan has requested a way to give Customers a text message on their phone after automatically after the mark a Repair ticket as Completed in the portal.

### 1. Refactor `AwsS3Config` to a general `AwsConfig`

This can then create beans for S3 and SNS respectively, for instance

```java
@Bean
    public AmazonS3 amazonS3(AWSCredentials credentials) {
        return AmazonS3ClientBuilder.standard()
                .withCredentials(new AWSStaticCredentialsProvider(credentials))
                .withRegion(region)
                .build();
    }

    @Bean
    public AmazonSNS amazonSNS(AWSCredentials credentials) {
        return AmazonSNSClientBuilder.standard()
                .withCredentials(new AWSStaticCredentialsProvider(credentials))
                .withRegion(region)
                .build();
    }
```

Also rename environment variables for aws access key, secret key and region to have S3 in their names.

This refactor means renaming in the java code but also in the pipelines where vars are used and in the `.env.template` - just you project search and replace all instances...

### 3. Use the sns service method in `TicketServiceImpl.updateTicketStatus()`

If is repair ticket and contains customer owned vehicles send message to customer on that ticket.

The message should be something like:

```
Din cykel är redo att hämtas på Hjulverkstan i {{shopName}}.
```

Where the shop name is taken from the location of the first vehicle in the ticket. Add a comment in code for this solution and point to a ticket that will talk further about this issue [link]

## Frontend

Add confirmation dialog when closing a repair ticket with customer owned vehicles:

Closing this ticket will send an sms to the customer.

Are you sure you want to close this ticket?
[cancel] [yes]

Look at delete confirmation dialog and how it is implemented

### LINKED ISSUE #296: Add location to ticket
Currently the location a ticket relates to is based on the vehicles it contains, but it can contain vehicles with multiple different locations, is that really decided?

When developing #267 we've realised this, as the location of the vehicles being repair should be used in the notification message sent upon completed repair for a customer.

The idea is for location to be a field on a ticket, and this has to be selected first before selecting the vechiles for a ticket. The vehicles dropdown will then only show vehicles in that location.

## Frontend

- Add a location before vehicle on ticket in the ShopTicketsFields.tsx (and data/tickets/(types.ts and form.ts)
- Vehicles dropdown should be disabled until a location is chosen.
- The list of vehicles should be filtered based on the selected location, do this by adding location as a prop to the `useTicketsAsEnumsQ` so that it can filter internally.
- `useTicketsAggregatedQ` currently creates a locationIds array on each ticket which is used for filtering and in the table. Remove this and use the locationId instead.

## Backend

- Add the location field as required on tickets

fendi86:


Cwejman:
@Jessica-a11y @andreljungsvik @MarioGit24 Lets review this next week.

Cwejman:
Rebased of main and slightly refactored.

Scope increased though as we need to keep track of sent message before this is production ready. See added scope in the issue #276

PR #306, opened 2025-06-25
Brute force test using python

karlojoseph:
-Added a pytest to perform brute force based on the common credentials found in the internet; ensuring that the correct authentication logic is in place. -Modifying dev web script to properly load the .env file during local run. 
-Updated .env template with new variables for pen testing


karlojoseph:
reopening

PR #309, opened 2025-07-03
Implementation of Smoke test in pipeline

karlojoseph:
* Added a smoke test suite for Hjulversktan portal. 
* Add 2-retry mechanism to smoke tests to handle occasional failures from fast test execution and flaky environments.
* added smoke test job in the pipeline to run smoke tests whenever there are frontend and/or backend changes.


Cwejman:
Apologies for the delay,

Nicely done, I'll review more during the week. Lets get it in as soon as possible

web/package.json
Cwejman: Better to add a script here in the package json called `test`, that way the pipeline, and local test does not need to execute all the steps for playwright.

web/tests/pages/LoginPage.ts
```diff
@@ -0,0 +1,59 @@
+import { Page, expect } from '@playwright/test';
+
+export class LoginPage {
+  private readonly baseUrl: string;
+
+  constructor(private page: Page) {
+    // Default to development URL if environment variable is not set
+    this.baseUrl = process.env.VITE_FRONTEND_URL || 'https://dev.hjulverkstan.org';
```
Cwejman: I think it is best for the env vars used in test to be prefixed with TEST_ to decouple them, like the python tests are implemented.

web/tests/pages/LoginPage.ts
```diff
@@ -0,0 +1,59 @@
...
+  private readonly usernameInput = this.page.getByRole('textbox', { name: 'username' });
+  private readonly passwordInput = this.page.getByRole('textbox', { name: 'password' });
+  private readonly loginButton = this.page.getByRole('button', { name: 'Login' });
+  private readonly submitButton = this.page.locator('#submit');
+  private readonly errorNotification = this.page.locator('[role="status"]', { hasText: 'Failed to log in' }).first();
+
+  // Actions
+  async goto() {
+    // Remove any leading slash from the path to avoid double slashes
+    const path = 'sv/portal'.replace(/^\/+/, '');
```
Cwejman: Here the url is swedish then but the tests are for English texts, the public site is localised but the portal thought isn't actually yet, but will some day, propose to use en instead then so that the tests still work when localisation is fully implemented.

web/tests/smoke/HjulverkstanPortalSmoke.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    await loginPage.login('invalid', 'invalid');
+    await loginPage.verifyLoginError();
+    
+    // Verify the full error message is shown after login attempt
+    const errorMessage = await loginPage.getErrorMessage();
+    expect(errorMessage).toContain('Try again soon or contact your local developer');
+  });
+
+  test('should login successfully with valid credentials', async () => {
+    await loginPage.login('admin', 'password');
```
Cwejman: Perhaps these credentials needs to be env vars?

As we dont want to expose the username and password....

Cwejman (changes requested):
Left a few comments and questions.

If you dont have time to work further on this that is of course fine.

I'll give you some time to answer, otherwise I'll close the issue as I do not have time finish the work myself.

Issue #312, opened 2025-09-10
Website vehicle detail page


Cwejman:
https://www.figma.com/design/xAkgj6iKQ2cEqgHnG5cxKI/Hjulverkstan?node-id=1461-8870

MarioGit24:
- Images: No images available at the moment.

- Mobile view (Bike Details – Bike Information): Content is too tightly packed, needs more spacing to align with the Figma example.

- Tablet: Layout looks good.

- iPad Pro (Bike Details): Text is positioned to the left of the Rent link instead of above.

PR #313, opened 2025-09-19
feat(web): VehicleDetail page

andreljungsvik:
Add vehicle detail page.

Solves #312

### DIFF
--- a/web/src/root/VehicleDetail.tsx
+++ b/web/src/root/VehicleDetail.tsx
@@ -1,0 +1,274 @@
+import { useMemo, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import {
+  Bike,
+  Cog,
+  Disc3,
+  Flag,
+  KeyRound,
+  LucideIcon,
+  Ruler,
+  Tag,
+} from 'lucide-react';
+
+import { Page } from '@components/Page';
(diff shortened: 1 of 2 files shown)


MarioGit24:
- Images: No images available at the moment i guess.

- Mobile view (Bike Details – Bike Information): Content is too tightly packed, needs more spacing to align with the Figma example.

- Tablet: Layout looks good.

- iPad Pro (Bike Details): Text is positioned to the left of the Rent link instead of above.

Cwejman:
Also resolve the images not working in dev environment so that we can fully test. I can support here if needed

Cwejman (changes requested):
Visual review:

1. Rent button and "Vi erbjuder uthyrning enbart i verkstaden" should vertically center aligned in its parent, button is of center.
2. Closed button and "Tillgänglig I Backa" should not be space between on 900px width, but have the same margin as other side of info div.
3. Closed/open badge should have white background
4. Margin top and bottom of attributes section should be same as margin between its rows
5. Dashed line between "Tillgänglig I Backa" and "Vi erbjuder uthyrning enbart i verkstaden" sections should be full width as in design.

MarioGit24 (approved):

Issue #319, opened 2025-10-02
Navigation does not reset scroll


andreljungsvik:
When navigating between routes the scroll position remains at the previous page. We need implement a ScrollToTop behavior to reset scroll on route changes.

PR #320, opened 2025-10-02
fix(web): Pages now scroll to top when navigating between routes

andreljungsvik:
Solves #319


Jessica-a11y (approved):
Reviewed the code, nothing to comment.

Issue #226, opened 2025-04-03
Website localisation for non-webedit text


Cwejman:
Essentially storing language in JSONs or the like and some kind of solution to read from it using the language context. Maybe a localisation library is needed for easier pluralis, singularis, counting etc... instead of doing everything our selves.

But I think opting in for a large react based library for i18n is not th way to go, we already have a LanguageContext for which is the selected language, so using a localisation library together with the language context in hook to just access localised strings just like usePreloadedData does is a development friendly and simple solution.

An extract from my brainstorming with ChatGPT on finding a good library:

## Prompt

[...]

PR #322, opened 2025-10-03
feat(web): Add website localisation for non-webedit text

andreljungsvik:
Solves #226


Jessica-a11y (approved):

web/src/components/OpenBadge.tsx
```diff
@@ -30,6 +31,8 @@ export const OpenBadge: React.FC<OpenBadgeProps> = ({
     />
   );
 
+  const { t } = useTranslations();
```
Cwejman, line 34: useTranslation should be at top of component, both because it looks better but also because hooks may not be conditionally invoked (it is used after a if() return null)...

web/src/components/PageNavbar.tsx
```diff
@@ -5,23 +5,25 @@ import { IconButton, Link } from '@components/shadcn/Button';
 import { useScrolledPastElement } from '@hooks/useScrolledPastElement';
 import { cn } from '@utils';
 
+import { useTranslations } from '@hooks/useTranslations';
+
 export const navLinks = [
-  { name: 'Home', path: '/' },
+  { key: 'home', path: '/' },
```
Cwejman, line 11: Better to call the key localisationKey since we do not invoke t('') here we do not know the purpose of the key

web/src/data/translations/en.ts
```diff
@@ -0,0 +1,79 @@
...
+  storiesLinkLabel: 'See all stories',
+  shopsLinkLabel: 'See all shops',
+
+  joinUsHeading: 'Join us',
+  contactUsLinkLabel: 'Contact us',
+
+  partnerHeading: 'Our partners',
+  partnerLinkLabel: 'See all partners',
+
+  footerSCBody: 'initiative by',
```
Cwejman, line 36: SC not quite understandable, perhaps should be SaveTheChildren

web/src/data/translations/sv.ts
```diff
@@ -0,0 +1,79 @@
...
+  contactUsLinkLabel: 'Kontakta oss',
+
+  partnerHeading: 'Våra partners',
+  partnerLinkLabel: 'Se alla partners',
+
+  footerSCBody: 'initiativ av',
+  footerSCLinkLabel: 'Rädda Barnen',
+  footerDevelopBody: 'utvecklat av',
+  footerDevelopLinkLabel: 'ALTEN Sverige',
+  footerOpenSourceBody: 'öppen källkod på',
```
Cwejman, line 40: 😂 perhaps "open source" is better than "öppen källkod"

web/src/hooks/useTranslations.ts
```diff
@@ -0,0 +1,31 @@
...
+import { translations, type TranslationKeys } from '@data/translations';
+import { fallBackLocale } from '@root';
+
+export type { TranslationKeys } from '@data/translations';
+
+export function useTranslations() {
+  const { currLocale } = usePreloadedDataLocalized();
+
+  const strings = useMemo(() => {
+    return (
```
Cwejman, line 13: No need to block scope with return, it is an arrow function, return directly

web/src/hooks/useTranslations.ts
```diff
@@ -0,0 +1,31 @@
...
+      translations[currLocale as keyof typeof translations] ?? fallBackLocale
+    );
+  }, [currLocale]);
+
+  const t = <K extends TranslationKeys>(
+    key: K | undefined,
+    values?: Record<string, any>,
+  ): string => {
+    if (!key) return '';
```
Cwejman, line 22: Same here, the whole fn could be one ternary, if you feel it is readable. The guidelines states that we prefer ternaries and declarative style over imperative – as long as it feels readable. You decide.

web/src/root/index.tsx
```diff
@@ -77,54 +78,54 @@ export const createRoutes = (
...
     {
       path: '/support',
-      title: 'Hjulverkstan - Support',
+      key: 'supportTitle',
       component: Support,
     },
     {
       path: '/services',
-      title: 'Hjulverkstan - Services',
+      key: 'servicesTitle',
```
Cwejman, line 118: Either use t() directly here and keep the name as title or change the name to something like `titleTranslationKey`

web/src/components/PageNavbar.tsx
Cwejman: Perhaps translationKey is better

Cwejman (changes requested):
Also after resolving all changes and amending, run `npm run format` to apply all unapplied prettier formatting and commit this as a `style` commit.

web/src/components/WarningBadge.tsx
```diff
@@ -1,25 +1,37 @@
...
 
 export default function WarningBadge({
   id,
   warnings,
   variant,
 }: WarningBadgeProps) {
+  const { t } = useTranslations();
+
+  const warningFind = (value: string) =>
+    warningEnums.find((e) => e.value === value) ?? {
```
Cwejman: Why manual find and not using the helpers?

web/src/root/VehicleDetail.tsx
```diff
@@ -109,11 +111,14 @@ export default function VehicleDetail() {
...
 
+  const { t } = useTranslations();
+  const enums = useTranslateRawEnums(enumsRaw);
+
   const [visibleCount, setVisibleCount] = useState(ITEMS_TO_LOAD_OTHER_BIKES);
 
   const vehicleQ = usePublicVehicleByIdQ({ id });
   const vehiclesQ = usePublicVehiclesByLocationQ({
-    locationId: vehicleQ.data?.locationId,
+    locationId: vehicleQ.data?.locationId ?? '',
```
Cwejman: Sus

Cwejman (dismissed):

MarioGit24 (approved):

PR #326, opened 2025-10-06
Complete backend overhaul with accompaning changes in frontend and extras.

Cwejman:
Hello, my apologies for the absurd commit. It really would of been best to se a more gradual and structured approach. However, this was the course of which I fell into the rabit hole, reconstructing everything at once...

Okay, what has happened?

- Me moved to a feature based structure.
- Unnecessary interface files removed in favour of smaller footprint.
- DTOs unified to one with mapping migrated into it
- Controllers where refined, giving same order of methods and removing response entity in favour of direct return with accompanying @ResponseStatus (see readme)
- Transactions were refined (some wrong imports) and strict in terms of read and write
- Manual constructor for injection replaced with @requiredargsconstructor
- Validation moved to Utils where complex validation rules are required.
- Services now follow a strict design pattern (see the readme)
- General Content renamed to Text
- Utils made for web edit to simplify all its services
- Security layer rewritten to me more aligned with best practices, trimming unnecessary footprint.
- Error handling in advice controller more coherent and no missing logs or stack traces
- Front ends src/data directory cleaned up and made more coherent, does do as much heavy lifting as some logic is moved to the backend (id as strings not long) webedit built an object instead of manual construction in api layer.

Maybe that is it?

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -1,5 +1,20 @@
 # Guidelines
 
+## Table of contents `📖`
+
+* [Guidelines](#guidelines)
+  * [Table of contents `📖`](#table-of-contents-)
+  * [Principles `⛩️`](#principles-)
+    * [Simplicity and coherence `🍇`](#simplicity-and-coherence-)
+    * [Describe over instruct `🫵`](#describe-over-instruct-)
+    * [Composition `🧱`](#composition-)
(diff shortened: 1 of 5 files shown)


Cwejman:
@MarioGit24 has approved but forgot to press the button. Merging anyway.

PR #338, opened 2025-10-22
(feature): Website language switcher

MarioGit24:


andreljungsvik:
The language switcher jumps around on small screens (and disappears off the screen on medium).  Maybe just keep the globe icon and make the dropdown a bit narrower on small screens.

web/src/components/PageNavbar.tsx
```diff
@@ -28,20 +34,29 @@ export interface PageNavbarProps {
...
+
   const [isOpen, setIsOpen] = useState(false);
   const scrolledPast = hasHeroSection
     ? useScrolledPastElement('section', 71)
     : false;
+
   const onHero = !scrolledPast && hasHeroSection && !isOpen;
 
+  const handleLangSelect = (locale: LangSlug) => {
+    navigate('/' + locale, { replace: true });
```
Cwejman: Does this work on other pages than home?

I would think that if you are on for instance `/en/vehicle/:id` and change to `se`  then you'd want the new path to be `/se/vehicle:id`

Cwejman:
Just one comment then, deployed to dev and looks good. Will also need to be rebased with main before "merging" is possible

Issue #291, opened 2025-05-26
Website story detail page


Cwejman:
[Story Detail](https://www.figma.com/proto/xAkgj6iKQ2cEqgHnG5cxKI/Hjulverkstan?node-id=1513-12503&t=6uKKTKcxviP56dWT-1&scaling=min-zoom&content-scaling=fixed&page-id=1%3A3&starting-point-node-id=1513%3A12503&show-proto-sidebar=1)

Needs a refined approach to parsing and rendering rich text - either from markdown or html.

PR #339, opened 2025-10-22
feat(web): Story detail page

Jessica-a11y:
Solves #291


Cwejman (changes requested):
Discussed in person

web/src/root/VehicleDetail.tsx
```diff
@@ -143,16 +143,17 @@ export default function VehicleDetail() {
     setVisibleCount((prevCount) => prevCount + ITEMS_TO_LOAD_OTHER_BIKES);
 
   return (
-    <Page variant="muted">
+    <Page
+      heading={
+        vehicleQ.data
+          ? `${findEnum(enums, vehicleQ.data.brand)?.label} ${findEnum(enums, vehicleQ.data.vehicleType)?.label}`
+          : 'A vehicle'
```
Cwejman: This results in "A vehicle" while the data is loading. I think leaving it empty like is was is better UX, best would be to use skeleton component but that is a bit premium for now. "A vehicle" is also not localised

web/src/components/TiptapContent.tsx
```diff
@@ -0,0 +1,23 @@
...
+  content,
+  className,
+}) => {
+  const html = useMemo(
+    () => content && generateHTML(content, [StarterKit]),
+    [content],
+  );
+
+  return html ? (
+    <div className={className} dangerouslySetInnerHTML={{ __html: html }} />
```
Cwejman: I think the styling for tip tap content should always be the same?

The class could then be set internally rather that form props, but leaving className attachable from props as well for further config is of course.

Cwejman (changes requested):

Cwejman (changes requested):
I ran the migration script from your other branch (which perhaps should have been here but never mind...) and deployed the pipeline to dev. Tried it out and the title for the the stories are missing, both on the cards and the detail pages.

The margin above the image on less than `md` is also too little. And I think the margin between the text and the image could be less when bellow `md` as well. The large margin between image and text works well for the larger layout but ifeels a bit far on smaller screens.

Issue #283, opened 2025-05-23
Use run configurations as files and gitignore .idea


Cwejman:
Currently .idea is not gitignored properly, this happened when we pushed more changes to the default .idea.

With should still keep the default .idea so that new users get preferred configuration regarding, language, prettier etc. But storing the run configurations as file using the open "store as file" on Arun configuration would make the .idea directory more correct.

1. Store run configurations as file
2. Make sure .idea directory is gitignored

PR #424, opened 2026-03-06
chore: Clean up .idea folder and add shared run configurations

Jessica-a11y:
This commit untracks personal IntelliJ settings like workspace.xml and compiler.xml that were previously committed by mistake. It also moves project-specific run configurations to the .run directory. 
Solves #283


MarioGit24 (approved):

andreljungsvik (changes requested):
The project has lost its 'Source Root' definition. IntelliJ no longer recognizes the src folder as Java source code, which means we can't run MainApplication at all (the IDE treats it as a plain text file).

PR #427, opened 2026-03-09
Feature/simplified webedit translations

### LINKED ISSUE #391: WebEdit simplified translations and CRUD
As the #331 was moved from idea to starting to be implemented we realised that the current contract of working with the web edit entities over CRUD with the global mode is not quite right.

Lets remove the global mode and the silos of working with entity OR working with the translatable fields. Instead you always CRUD for a language but there is a set default language that is mandatory and you can only publish if a translation exists for this language.

Read and update is free, but naturally bound by language, read needs two separate ends met, to read with the default language as fallback or reading purely only for the selected language.

Delete is interesting, soft deletes will be implemented with #331 but without thinking about the release flow with drafts, archiving and everything one should be able to delete per language, though not be able to delete the default language, rather then, delete the whole entity. For now before releases/draft logic is in place, lets keep it simple, delete deletes everything.

Also preparing for #331 let's rework localised_content a bit. Lets connect the entities instead to a `translation_set`, this has an id and the `WebEditEntity`, the localised_content, renamed to `translation` then just references that `translation_set` and doesn't care about the entities or their ids.

Cwejman:


Jessica-a11y (changes requested):
When creating a story, neither the title or body text is saved in the translation_set. It does get an identity id though.

Jessica-a11y (approved):

andreljungsvik (approved):

PR #457, opened 2026-03-16
feat: Add frame number to Vehicle

### LINKED ISSUE #131: Add support for framenumber
Hjulverkstan is a trust-based organisation and the surrounding society in different ways helps and support the operation. 

There may be reasons in the future to ask hjulverkstan regarding bike(s).
If hjulverkstan does not have support for framenumber (ramnummer) in the system it might look like the organization has a laisse-faire approach to the moral rules in the surrounding society in case there are misuse being conducted.

Example use case that should not be read literally but serves as an example of the risk of having its own separate anonymous bubble:
One or many children are being used to get bikes that is later sold and older bbad guys collects the money for this.
The bikes are later included in an investigation and the kid tells that he got the bikes from hjulverkstan.
Hjulverkstan is not able to answer questions regarding the bikes.
Hjulverkstan is now helping the bad guys and being cruel to the kids instead of the opposite intended action and looses credibility and trust. The bad guys like hjulverkstan, it gives them a chance to hook up kids and then use them for other more sinister actions. 

It also gives people inside hjulverkstan with bad intent to stay under the radar since the bikes apply to fire and forget pattern.

andreljungsvik:


Jessica-a11y (approved):
I suggest we talk about implement conditional validation for the frame number field. For bikes owned by the company, the frame number should be mandatory to ensure asset traceability and insurance compliance. However, for customer-owned bikes, we can keep it hidden.

PR #480, opened 2026-03-25
feat: implement user location awareness and form auto-fill

### LINKED ISSUE #434: Add Location on Customer
Right now all of the shops can see each others customers. This will create a long list (maybe of duplicates) customers each day and make it harder to find the right one. Could we put Location on the Customer entity to make this easier?

### LINKED ISSUE #436: Default Location in Portal
For easier naviagtion, creation and filter, the Portal could have Location default depending on which user is logged in. 
They should also be able to switch between seeing all vehicles from all shops adn just seeing their own.

### LINKED ISSUE #437: Default Location on Vehicle
To lessen the work of creating a Vehicle, Location could be default depending on which Shop you're at.

### LINKED ISSUE #470: Assign Default Location Context based on Logged-in User
To streamline workflows and resolve issues with cross-shop data clutter (as seen in #434, #436, and #437), we need to tie data creation to the user's specific shop.

The idea is that each user will be assigned a specific location_id based on the shop they select when their account is created. Once logged in, the system will automatically fetch the user's location_id and set it as the default location whenever a new ticket, vehicle, or customer is created.

In other words, selecting a shop will be mandatory when creating a new user. Users will still be able to view vehicles and tickets from other shops if they want to, but the default location will always be based on the currently logged-in user's shop.

Closes #434 

Closes #436 

Closes #437

andreljungsvik:


Jessica-a11y (changes requested):
When creating an account for Frölunda, you can't see Frölunda in the location-filter dropdown in the Vehicle inventory. You can only unchheck locations that has vehicles.
There's a 'clear filter' option on location-filter, but nothing happens in the inventory.

Jessica-a11y (approved):
Great job! Good that location is optional!

PR #484, opened 2026-03-26
style(web): Fix ticket completion dialog layout

### LINKED ISSUE #443: Notification dialog
When the dialog for completing a ticket pops up, the dialog looks bad. Needs some margin or padding. Header is almost inside the X icon.

Jessica-a11y:
Improved the visual of the ticket completion dialog by adding more width. The header text was previously overlapping with the close icon. Solves #443


andreljungsvik (approved):
Great that the overlap is fixed! However, the new width feels a bit excessive and makes the layout look slightly unbalanced. It is quite a minor detail though, and it looks OK now!

PR #488, opened 2026-03-30
Anonymize customers upon removal & Validation of opening hours.

MarioGit24:
- Added validation for opening hours (HH:MM format)
  - Customers with active tickets are now anonymized instead of blocked from deletion, supporting GDPR purge requirements
  - Anonymized customers are automatically deleted when their last ticket is removed

### DIFF
--- a/api/scripts/361.sql
+++ b/api/scripts/361.sql
@@ -1,0 +1,1 @@
+ALTER TABLE customer ADD COLUMN anonymized BOOLEAN NOT NULL DEFAULT false;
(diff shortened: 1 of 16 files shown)


andreljungsvik (approved):

Jessica-a11y (changes requested):
If a Employee is connected to closed tickets, it cannot be GDPR-purged. The delete button is disabled.
When removing a Employee with no tickets attached to it, the delete message should say its a purge and that emplyee will be permanently removed. 

A better delete message for user is also needed. 

We also need a better message for permanently deleting a customer.

PR #501, opened 2026-04-14
feat(web): New logo

### LINKED ISSUE #493: Add new Logo
Add the new logo and make it center on the Home page

Jessica-a11y:
Solves #493


andreljungsvik (changes requested):
We should use an SVG instead of a PNG for the logo. It would ensure the logo stays crisp on all screen types and keeps the file size down.

andreljungsvik (approved):

Issue #539, opened 2026-04-29
ImageCard navigation


Jessica-a11y:
The whole card should act as a link. You should be able to click the card to open the dialog.

PR #556, opened 2026-04-30
style(web): ImageCard hover shadow and clickable

Jessica-a11y:
Increased shadow on hover for improved visual feedback and made ImageCard clickable.

Solves #539


MarioGit24 (changes requested):
you have several duplicate entries on data.sql
  - c3a00000-...000000000026 — two 'SV' rows, no 'EN' row
  - c3a00000-...00000000001a — two 'EN' rows, no 'SV' row
  - c3a00000-...00000000001e — two 'EN' rows, no 'SV' row
  - c3a00000-...00000000001f — two 'SV' rows, no 'EN' row
  - c3a00000-...00000000002a — two 'SV' rows, no 'EN' row
  - c3a00000-...00000000002b — two 'SV' rows, no 'EN' row
  - c3a00000-...00000000002e — two 'SV' rows, no 'EN' row
  For example. 
This is a bug. and could result in duplicate rows, and key constraint violations. 

This happened when i tested it myself: 
1. Missing language → NULL foreign key: When a translation_set row is missing for a language (e.g., 000000000026 has no EN row), the SELECT subquery returns 0 rows. PostgreSQL inserts NULL as the translation_set_id, creating an orphaned translation with no link to any translation set. That translation    
  will never be found by the app.                                                                                                                                                                                                                                                                                   
  2. Duplicate language → hard crash: When a translation_set has two rows for the same language (e.g., 00000000002a has two SV rows), the SELECT subquery returns 2 rows and PostgreSQL throws:                                                                                                                     
  ERROR: more than one row returned by a subquery used as an expression                                                                                                                                                                                                                                             
  2. This will fail the entire seed and prevent the app from starting. 


 What's Good

  - The link -> to rename is a nice consistency improvement.
  - The shadow-card-glow-hover tailwind extension is clean — the layered box-shadow values give a nice visual feedback effect.

